PR #27, opened 2021-03-24
remove dockerfile.prow

leena-jawale:


leena-jawale:
https://github.com/open-cluster-management/managedcluster-import-controller/pull/27

PR #28, opened 2021-03-25
Remove SyncSet in favor of Applier

itdove:
https://github.com/open-cluster-management/backlog/issues/6468
https://github.com/open-cluster-management/backlog/issues/10576


pkg/controller/managedcluster/managedcluster_import_detach.go
```diff
@@ -23,35 +23,69 @@ import (
...
 
 func (r *ReconcileManagedCluster) importCluster(
 	managedCluster *clusterv1.ManagedCluster,
+	clusterDeployment *hivev1.ClusterDeployment,
 	autoImportSecret *corev1.Secret) (res reconcile.Result, err error) {
 	res = reconcile.Result{}
-	var client client.Client
+
+	//Assuming that is a local import
+	client := r.client
```
hanqiuzh, line 37: does that mean we will always import local? even for manual import scenario?

itdove: The importCluster is called only if clusterDeployment or autoImportSecret present or managedCluster has the label for local-import.
https://github.com/itdove/managedcluster-import-controller/blob/05910be5a8ff93d4c3[…]216d/pkg/controller/managedcluster/managedcluster_controller.go
Line 292

PR #31, opened 2021-03-30
Quotes for name and namespace

chrisahl:
Forward port for issue https://github.com/open-cluster-management/backlog/issues/10273


resources/klusterlet/klusterlet.yaml
```diff
@@ -5,10 +5,10 @@ kind: Klusterlet
 metadata:
   name: klusterlet
 spec:
-  registrationImagePullSpec: {{ .RegistrationImageName }}
-  workImagePullSpec: {{ .WorkImageName }}
-  clusterName: {{ .ManagedClusterNamespace }}
-  namespace: {{ .KlusterletNamespace }}
+  registrationImagePullSpec: "{{ .RegistrationImageName }}"
+  workImagePullSpec: "{{ .WorkImageName }}"
```
leena-jawale, line 9: do we need quotes for images ?

chrisahl: I could probably pull those out if there is not much of them having weird values

chrisahl: changed

PR #33, opened 2021-03-31
Add check if pods in ns before delete

itdove:

### DIFF
--- a/pkg/controller/managedcluster/managedcluster_controller.go
+++ b/pkg/controller/managedcluster/managedcluster_controller.go
@@ -444,6 +444,19 @@ func (r *ReconcileManagedCluster) deleteNamespace(namespaceName string) error {
 			namespaceName,
 		)
 	}
+
+	if tobeDeleted {
+		pods := &corev1.PodList{}
+		err = r.client.List(context.TODO(), pods, client.InNamespace(namespaceName))
+		if err != nil {
+			return err
+		}
+		if len(pods.Items) != 0 {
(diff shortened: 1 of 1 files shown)


leena-jawale:
There is namespace `openahift-config` which does not have pod. In that case, it will delete that namespace

hanqiuzh:
agree with @leena-jawale , we may want another rule like regex `openshift-*`?

jnpacker:
I wasn't sure how far to go. But this would stop options like openshift-josh1 and openshift-east-cluster01

jnpacker:
But, this just means we won't delete the namespace so I can get onboard with openshift-*. Just make sure it gets documented.

itdove:
Guys, this a first step until they accept an overall cr containing all information to start an import... as we proposed few months ago !
Pleat just accept this change as a temporary solution !

hanqiuzh:
Will approve this pr for now. Most openshift namespaces seems are either protected by webhook, or it can generate itself.
```
# oc delete ns openshift-config
Error from server (Forbidden): namespaces "openshift-config" is forbidden: this namespace may not be deleted
# oc delete ns openshift-dns
namespace "openshift-dns" deleted
# oc get ns openshift-dns
NAME            STATUS   AGE
openshift-dns   Active   22s
//!!!!!
# oc delete ns openshift-kube-apiserver 
namespace "openshift-kube-apiserver" deleted
# oc get ns openshift-kube-apiserver 
NAME                       STATUS   AGE
openshift-kube-apiserver   Active   32s
```

PR #53, opened 2021-05-17
Add CSR delete/generic predicate

itdove:

### DIFF
--- a/pkg/controller/csr/csr_manager.go
+++ b/pkg/controller/csr/csr_manager.go
@@ -38,6 +38,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	csrPredicateFuncs := predicate.Funcs{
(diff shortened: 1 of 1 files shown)


pkg/controller/csr/csr_manager.go
```diff
@@ -38,6 +38,8 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	csrPredicateFuncs := predicate.Funcs{
+		GenericFunc: func(e event.GenericEvent) bool { return false },
```
hanqiuzh, line 41: do we want to disable generic? (e.g. pod crashed/restarted when csr is created, so we missed the create event?)

hanqiuzh: I think that edge case is ok, and this is same predicate as klusterlet-addon-controller 👍

PR #61, opened 2021-06-24
Add both crd on mw

itdove:
https://github.com/open-cluster-management/backlog/issues/13631
/hold

### DIFF
--- a/pkg/controller/managedcluster/managedcluster_controller.go
+++ b/pkg/controller/managedcluster/managedcluster_controller.go
@@ -335,6 +335,9 @@ func (r *ReconcileManagedCluster) Reconcile(request reconcile.Request) (reconcil
 			return reconcile.Result{Requeue: true, RequeueAfter: 30 * time.Second}, err
 		}
 		if isV1 {
+			// The 2 versions of the crd are added to avoid the unexpected removal during the work-agent upgrade.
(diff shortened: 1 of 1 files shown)


pkg/controller/managedcluster/managedcluster_controller.go
```diff
@@ -335,7 +335,8 @@ func (r *ReconcileManagedCluster) Reconcile(request reconcile.Request) (reconcil
 			return reconcile.Result{Requeue: true, RequeueAfter: 30 * time.Second}, err
 		}
 		if isV1 {
-			_, _, err = createOrUpdateManifestWorks(r.client, r.scheme, instance, crds["v1"], yamls)
+			ucrds := append(crds["v1"], crds["v1beta1"]...)
```
qiujian16, line 338: I would add a Note here indicate that we add this to avoid unexpected crd removal during work-agent upgrade, and we will remove v1beta1 in the next release or zstream.

itdove: Are people supposed to go through all z release... can't they jump from one release for example 2.3.0 to 2.3.4 in one shot? I will say we will remove it in a future z release.

itdove: comment added

PR #66, opened 2021-08-03
override images by annotaions

zhiweiyin318:
story: https://github.com/open-cluster-management/backlog/issues/14686


pkg/controller/managedcluster/import_yamls.go
```diff
@@ -74,9 +78,20 @@ func generateImportYAMLs(
 		return nil, nil, err
 	}
 
+	pullSecretNamespace := ""
+	pullSecret := ""
+	registry := ""
+	if managedCluster.Labels[clusterImageRegistryLabel] != "" {
+		registry, pullSecretNamespace, pullSecret, err = getImageRegistryAndPullSecret(client,
```
qiujian16, line 85: I would have one func and return secret here.

pkg/controller/managedcluster/import_yamls.go
```diff
@@ -147,21 +162,48 @@ func generateImportYAMLs(
...
+	defaultImage := os.Getenv(envName)
+	if defaultImage == "" {
+		klog.Errorf("the default image %v is invalid", envName)
+		return "", fmt.Errorf(envVarNotDefined, registrationImageEnvVarName)
+	}
+	if registry == "" {
+		return defaultImage, nil
+	}
+
+	registry = strings.TrimSuffix(registry, "/")
```
qiujian16, line 200: add some doc here.

PR #71, opened 2021-08-27
refactor - add scaffold methods

skeeey:
Signed-off-by: liuwei <liuweixa@redhat.com>


pkg/helpers/helpers.go
```diff
@@ -0,0 +1,641 @@
...
+	mapper, err := apiutil.NewDiscoveryRESTMapper(clientConfig)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return runtimeClient, mapper, nil
+}
+
+// AddManagedClusterFinalizer add a finalizer to a managed cluster
+func AddManagedClusterFinalizer(client client.Client, recorder record.EventRecorder,
```
qiujian16, line 169: if it is using controller-runtime, there is a util func here  https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L352

pkg/helpers/helpers.go
```diff
@@ -0,0 +1,641 @@
...
+			errs = append(errs, applyManifestWork(client, recorder, required))
+		case *operatorv1.Klusterlet:
+			errs = append(errs, applyKlusterlet(client, recorder, required))
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func applyServiceAccount(client client.Client, recorder record.EventRecorder, required *corev1.ServiceAccount) error {
```
qiujian16, line 386: why we need these apply funcs, couldn't we use library-go?

pkg/helpers/helpers.go
```diff
@@ -0,0 +1,641 @@
...
+		}
+
+		reportEvent(recorder, required, "Deployment", "created")
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	if equality.Semantic.DeepEqual(existing.Spec, required.Spec) {
```
qiujian16, line 505: this would be hard  with defaulting.

pkg/helpers/helpers.go
skeeey: will use them

pkg/helpers/helpers.go
skeeey: using library-go, we need adapt the `library-go` argument types with controller-runtime, but I can try use library-go

pkg/helpers/helpers.go
qiujian16: do you need an event when adding a finalizer?

skeeey: I hope to record an event when I add/remove a finalizer from managed cluster

pkg/helpers/helpers.go
```diff
@@ -0,0 +1,641 @@
...
+		if err != nil {
+			klog.Warningf("The value of %s env is wrong, using default reconciles (1)", maxConcurrentReconcilesEnvVarName)
+			maxConcurrentReconciles = 1
+		}
+	}
+	return maxConcurrentReconciles
+}
+
+// SetLabels set a label on the given object
+func SetLabels(objMeta *metav1.ObjectMeta, key, val string) bool {
```
zhiweiyin318, line 83: suggest to refer this method https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourcemerge/object_merger.go#L138  can update/add/remove labels.

pkg/helpers/helpers.go
```diff
@@ -0,0 +1,641 @@
...
+func AddManagedClusterFinalizer(client client.Client, recorder record.EventRecorder,
+	managedCluster *clusterv1.ManagedCluster, finalizer string) error {
+	for i := range managedCluster.Finalizers {
+		if managedCluster.Finalizers[i] == finalizer {
+			return nil
+		}
+	}
+
+	managedCluster.Finalizers = append(managedCluster.Finalizers, finalizer)
+	if err := client.Update(context.TODO(), managedCluster); err != nil {
```
zhiweiyin318, line 178: should we consider retry here if fails?  or write another update function with retry?

pkg/helpers/helpers.go
```diff
@@ -0,0 +1,641 @@
...
+		}
+		copiedFinalizers = append(copiedFinalizers, managedCluster.Finalizers[i])
+	}
+
+	if len(managedCluster.Finalizers) == len(copiedFinalizers) {
+		return nil
+	}
+
+	managedCluster.Finalizers = copiedFinalizers
+	if err := client.Update(context.TODO(), managedCluster); err != nil {
```
zhiweiyin318, line 204: should we consider retry here if fails? or write another update function with retry?

pkg/helpers/helpers.go
skeeey: Yeah, after I imported the `library-go`, I will use it

pkg/helpers/helpers.go
skeeey: retry here may be not necessary, if this method return error, the controller will reconcile the request, it will equal retry

pkg/helpers/helpers.go
skeeey: retry here may be not necessary, if this method return error, the controller will reconcile the request, it will equal retry

pkg/helpers/helpers.go
skeeey: remove this two method, will use MergeMap directly

pkg/helpers/helpers.go
skeeey: using library-go to apply deployment

qiujian16 (commented):
/lgtm

PR #84, opened 2021-09-17
add nodeSelector to the pods of klusterlet

zhiweiyin318:
story: https://github.com/open-cluster-management/backlog/issues/15911


pkg/helpers/helpers.go
```diff
@@ -505,3 +509,18 @@ func GetComponentNamespace() (string, error) {
 	}
 	return string(nsBytes), nil
 }
+
+func GetNodeSelector(cluster *clusterv1.ManagedCluster) (map[string]string, error) {
+	// This part is to support running pods related local-cluster on specified nodes,like infra nodes.
+	nodeSelector := map[string]string{}
+	if cluster.GetName() == "local-cluster" {
+		annotations := cluster.GetAnnotations()
+		if nodeSelectorString, ok := annotations[AnnotationNodeSelector]; ok {
```
qiujian16, line 518: I think a better way is always short circuit.

``` go
if cluster.GetName() != "local-cluster" {
  return nodeSelector
}

annotations := cluster.GetAnnotations()
if annotations == nil {
   return nodeSelector
}

nodeSelectorString, ok := annotations[AnnotationNodeSelector]
if !ok {
  return nodeSelector
}
```

pkg/controller/importconfig/manifests/klusterlet/klusterlet.yaml
```diff
@@ -7,3 +7,12 @@ spec:
   workImagePullSpec: "{{ .WorkImageName }}"
   clusterName: "{{ .ManagedClusterNamespace }}"
   namespace: "{{ .KlusterletNamespace }}"
+  nodePlacement:
+    nodeSelector:
+    {{- range $key, $value := .NodeSelector }}
```
qiujian16, line 12: can NodeSelector be nil here?

pkg/helpers/helpers.go
zhiweiyin318: done

pkg/controller/importconfig/manifests/klusterlet/klusterlet.yaml
zhiweiyin318: nil is ok here, but NodeSelector must be defined in config.

PR #88, opened 2021-09-29
WIP: Customizing secret and manifestwork event sources

skeeey:
Signed-off-by: liuwei <liuweixa@redhat.com>


skeeey:
refer to https://github.com/open-cluster-management/backlog/issues/16595

pkg/source/secret_source.go
```diff
@@ -0,0 +1,147 @@
...
+	secretInformer cache.SharedIndexInformer
+}
+
+// NewImportSecretSource return a SecretSource only for import secrets
+func NewImportSecretSource(secretInformer cache.SharedIndexInformer) *SecretSource {
+	return &SecretSource{secretInformer: secretInformer}
+}
+
+// NewAutoImportSecretSource return a SecretSource only for auto import secrets
+func NewAutoImportSecretSource(secretInformer cache.SharedIndexInformer) *SecretSource {
```
qiujian16, line 37: It seems duplicate with NewImportSecretSource

zhujian7:
Can we hold this PR temporarily?
I find it seems there is a simpler way to implement this. I will try it and discuss it with @skeeey later.

pkg/source/secret_source.go
skeeey: Just use different name to distinguish different informer

zhujian7:
> Can we hold this PR temporarily? I find it seems there is a simpler way to implement this. I will try it and discuss it with @skeeey later.

Yesterday I found the [`SelectorsByObject option to cache`](https://github.com/kubernetes-sigs/controller-runtime/pull/1435) feature of the controller-runtime, I think we can set the label/field selector for `Secret` by overriding the `SelectorsByObject` while creating controller runtime manager. 
```
...
	// Create controller-runtime manager
	mgr, err := ctrl.NewManager(cfg, manager.Options{
		Scheme:             scheme,
		MetricsBindAddress: fmt.Sprintf(":%d", metricsPort),
		LeaderElection:     true,
		LeaderElectionID:   "managedcluster-import-controller.open-cluster-management.io",
		NewCache:           newCache,
	})
...
// newCache initializes and returns a new Cache.
func newCache(config *rest.Config, opts cache.Options) (cache.Cache, error) {
	labelSelector := labels.NewSelector()
	requirement, err := labels.NewRequirement(constants.ClusterImportSecretLabel, selection.Exists, nil)
	if err != nil {
		return nil, err
	}
	labelSelector = labelSelector.Add(*requirement)

	fieldSelector := fields.OneTermEqualSelector("metadata.name", constants.AutoImportSecretName)

	opts.SelectorsByObject = cache.SelectorsByObject{
		&corev1.Secret{}: {
			// Label and Field selector should be OR relationship instead of AND.
			Label: labelSelector,
			Field: fieldSelector,
		},
	}
	return cache.New(config, opts)
}

```

But we need the `auto import secret`(field selector, result A) and the `import secret`(label selector, result B), **union** of A and B. But `SelectorsByObject` feature gets the **intersection** of A and B. 

It is close to `code freeze` I think we can merge this PR now. And in the future, we can
- add a specific label to the `auto import secret` and make all secrets we are interested in could be selected by labels, then we can use `SelectorsByObject` to filter the `Secret` cache.

OR

- push the upstream controller-runtime to support **union** label selector and field selector of `SelectorsByObject option to cache`  

/hold cancel

PR #89, opened 2021-09-29
Add posthook manifestwork capability

itdove:
https://github.com/open-cluster-management/backlog/issues/15933

### DIFF
--- a/pkg/controller/manifestwork/manifestwork_controller.go
+++ b/pkg/controller/manifestwork/manifestwork_controller.go
@@ -6,7 +6,9 @@ package manifestwork
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/open-cluster-management/managedcluster-import-controller/pkg/constants"
 	"github.com/open-cluster-management/managedcluster-import-controller/pkg/helpers"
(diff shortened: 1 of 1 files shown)


pkg/controller/manifestwork/manifestwork_controller.go
```diff
@@ -256,6 +270,34 @@ func (r *ReconcileManifestWork) forceDeleteManifestWork(ctx context.Context, nam
...
+	manifestWorkNames := []string{}
+	for _, manifestWork := range manifestWorks.Items {
+		if gracePeriodString, ok := manifestWork.GetAnnotations()[posthookAnnotation]; ok {
+			gracePeriod, err := strconv.Atoi(gracePeriodString)
+			if err != nil {
+				return false, err
+			}
+			if time.Since(cluster.DeletionTimestamp.Time) < time.Duration(gracePeriod)*time.Second {
+				manifestWorkNames = append(manifestWorkNames, manifestWork.GetName())
+			}
```
skeeey, line 289: if the deletion time of the post hook works is grater than grace period, do we need force delete the works?

And the other question is do we handle the post hook works deletion in the post hook controller? 
I think it may be better to handle them in the post hook controller, if we handle the deletion in that controller, I think we don't need to check the post hook works in here, because if the post hook works are not deleted, they will can be found even if the works are in the deletion stage, we will not delete the managed cluster,  because there are still other works excepting klusterlet works

What's your opinion?

itdove: Hi, I invite you in the cloud-service team channel as I started a thread there with QiuJian about the post hook
5:33
My last comment is about put the annotation in the managedcluster rather than in the manifest work.
5:36
The main goal of the post hook is to avoid to delete the klusterlet before the post hook is executed as the klusterlet is needed in order to execute any manifestwork.
I think putting in a separate controller will not prevent the klusterlet to be removed.

itdove:
Need further investigation

PR #105, opened 2021-11-25
[WIP] auto-import hypershift guest(hosted) clusters:

ianzhang366:
1. watch hostedcluster CR
2. create managedcluster CR, namespace, klusterletaddonconfig upon reconciling

Note: ignore finalizer order for hostedcluster reconcile

Signed-off-by: Ian Zhang <izhang@redhat.com>


pkg/controller/hypershift/hostedcluster_controller.go
```diff
@@ -0,0 +1,387 @@
...
+
+	hostedClusterClient, restMapper, err := helpers.GenerateClientFromSecret(hostedKubeconfigSecret)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	importSecretName := fmt.Sprintf("%s-%s", clusterName, constants.ImportSecretNameSuffix)
+	importSecret, err := r.kubeClient.CoreV1().Secrets(clusterName).Get(ctx, importSecretName, metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		return reconcile.Result{}, nil
```
zhiweiyin318, line 179: I think this should return err to reconcile until the import-secret is created if we do not watch the import-secret.
otherwise there is no event to trigger reconcile since we only watch hostedCluster.

pkg/controller/hypershift/hostedcluster_controller.go
```diff
@@ -0,0 +1,387 @@
...
+		return reconcile.Result{}, err
+	}
+
+	// add a managed cluster finalizer to the hosted cluster, to handle the managed cluster detach case.
+	if err := r.addClusterImportFinalizer(ctx, hCluster); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// if there is an auto import secret in the managed cluster namespce, we will use the auto import secret to import the cluster
+	_, err = r.kubeClient.CoreV1().Secrets(clusterName).Get(ctx, constants.AutoImportSecretName, metav1.GetOptions{})
```
zhiweiyin318, line 148: I don't think we need auto-import secret in this scenario.  kubeconfig of hostedCluster.status is enough.

cmd/manager/main.go
```diff
@@ -127,10 +129,11 @@ func main() {
...
 	// Create controller-runtime manager
 	mgr, err := ctrl.NewManager(cfg, manager.Options{
-		Scheme:             scheme,
-		MetricsBindAddress: fmt.Sprintf(":%d", metricsPort),
-		LeaderElection:     true,
-		LeaderElectionID:   "managedcluster-import-controller.open-cluster-management.io",
+		Scheme:                  scheme,
+		MetricsBindAddress:      fmt.Sprintf(":%d", metricsPort),
+		LeaderElection:          true,
+		LeaderElectionNamespace: "open-cluster-management",
```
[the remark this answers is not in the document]
ianzhang366: Totally agree, this change is intended for running my local test

pkg/controller/hypershift/hostedcluster_controller.go
ianzhang366: for this one, I looked into `helpers.ImportManagedClusterFromSecret` which has checked for the import secret non-exist case:

```
	importSecret *corev1.Secret) error {
	if err := ValidateImportSecret(importSecret); err != nil {
		return err
	}
```

Do you think this is enough or not?

pkg/controller/hypershift/hostedcluster_controller.go
ianzhang366: I read the following:
https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.4/html/clusters/managing-your-clusters#importing-a-managed-cluster-with-the-cli

at section: `1.8.2.4. Importing the cluster`, it mentioned the auto-import secret stuff... do we need to worry about this for hypershift?

If not, we can remove the above section.

If so, we might want to have a logic to priorities, the auto-import secret and the auto-generated secret by `hostedCluster.Status.Kubeconfig`

zhiweiyin318: if user create a auto-import secret for the managedCluster, the autoimport controller will handle that, it is duplicate here.

pkg/controller/hypershift/hostedcluster_controller.go
zhiweiyin318: it returns nil and cannot go to `helpers.ImportManagedClusterFromSecret` if do not get import secret.

PR #106, opened 2021-12-08
fix issue #18006

skeeey:


skeeey:
/hold

util the issue #18006 is approved

skeeey:
refer to https://github.com/open-cluster-management/backlog/issues/18006

pkg/controller/managedcluster/managedcluster_controller.go
```diff
@@ -632,5 +645,20 @@ func isAPIExtensionV1(managedClusterClient client.Client,
...
+		return false, fmt.Errorf("kubernetes version not yet available for managed cluster %s", managedCluster.GetName())
+	}
+	actualVersion = managedCluster.Status.Version.Kubernetes
+	klog.V(4).Infof("actual kubernetes version is %s", actualVersion)
+	isV1, err := v1OnlyAPIExtensionMinVersion.Compare(actualVersion)
+	if err != nil {
+		return false, err
+	}
+	klog.V(4).Infof("isV1: %t", isV1 == -1)
+	return isV1 == -1, nil
```
elgnay, line 663: CRD v1beta1 is removed in kube 1.22. So `return isV1 == -1` or `return isV1 <= 0`?

skeeey: I think it should be `isV1 == -1`,  because
the result `-1` means the current kube version is less than 1.22.0, `0` means the current kube version equal to 1.22.0.
we should exclude the 1.22.0, am I right?

skeeey: fixed

PR #117, opened 2022-01-19
using claim as the cluster name label

skeeey:

### DIFF
--- a/pkg/controller/managedcluster/managedcluster_controller.go
+++ b/pkg/controller/managedcluster/managedcluster_controller.go
@@ -134,7 +134,11 @@ func (r *ReconcileManagedCluster) ensureManagedClusterMetaObj(ctx context.Contex
 	// if there is no cluster name label ensure the cluster name label
 	// TODO we should ensure only update the name label in one place
 	if name, ok := managedCluster.Labels[clusterNameLabel]; !ok || name == "" {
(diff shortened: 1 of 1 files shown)


skeeey:
refer to https://github.com/stolostron/backlog/issues/19229

pkg/controller/managedcluster/managedcluster_controller.go
```diff
@@ -232,6 +236,24 @@ func (r *ReconcileManagedCluster) deleteManagedClusterNamespace(
...
+	err := r.client.Get(ctx, types.NamespacedName{Namespace: cluster.Name, Name: cluster.Name}, clusterDeployment)
+	if errors.IsNotFound(err) {
+		return cluster.Name, nil
+	}
+
+	if err != nil {
+		return "", err
+	}
+
+	if clusterDeployment.Spec.ClusterPoolRef != nil {
```
ldpliu, line 250: You should also need check if clusterDeployment.Spec.ClusterPoolRef.ClaimName exist.

ldpliu, line 250: you may also need check if clusterDeployment.Spec.ClusterPoolRef.ClaimName is null

skeeey: I don't think we need check the claimname is null, because it's a string, can the ClaimName be empty?

ldpliu (approved):

PR #122, opened 2022-02-09
WIP: Import cluster in hypershift detached mode

zhujian7:
This PR supports importing a hypershift hosted cluster to the ACM hub in detached mode.
Users only need to create a managed cluster CR with
- label: import.open-cluster-management.io/klusterlet-deploy-mode: Hypershift-Detached
- annotation: import.open-cluster-management.io/management-cluster-name: cluster1
Then the import controller will start the importing process.

If users want to automatically import the hypershift hosted cluster, they are also required to provide an `auto-import-secret`. Or they can create the `external-managed-kubeconfig` on the management cluster manually.

And this feature can be controlled by a feature gate, it is disabled by default, you can enable it by adding a flag to the controller `--feature-gates=HypershiftImport=true`

Implement importing cluster in Detached mode according to the design https://github.com/stolostron/backlog/issues/18978#issuecomment-1021834313


zhujian7:
cc @elgnay

cmd/manager/main.go
```diff
@@ -64,7 +67,22 @@ func init() {
...
 
+// EnvironmentEnableHypershiftImportFG is an environment key for E2E test to enable
+// the HYPERSHIFT-IMPORT feature gate.
+const EnvironmentEnableHypershiftImportFG = "ENABLE_HYPERSHIFT_IMPORT"
+
 func main() {
+	pflag.CommandLine.SetNormalizeFunc(utilflag.WordSepNormalizeFunc)
+	features.DefaultMutableFeatureGate.AddFlag(pflag.CommandLine)
+	// E2E tests use env to tweak the feature-gates flag
+	if os.Getenv(EnvironmentEnableHypershiftImportFG) == "true" {
```
skeeey, line 78: I think we should set this flag on the e2e shell script, is there any special for this?

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -60,6 +71,24 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	mode, err := helpers.DetermineKlusterletMode(managedCluster)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
```
skeeey, line 77: I really think we should have a independent controller to handle the detached cluster :), for default managed cluster, we just return here

```
if mode == default {
  return nil
}
```

pkg/controller/importconfig/importconfig_controller.go
```diff
@@ -97,7 +96,17 @@ func (r *ReconcileImportConfig) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
-	// make sure the managed cluster clusterrole, cluserrolebiding and bootstrap sa are updated
+	mode, err := helpers.DetermineKlusterletMode(managedCluster)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	worker, ok := r.workers[mode]
```
skeeey, line 103: how about use a factory instead of this? we will get the worker from the factory

may be
```
worker, err := getImportWorker(mode)
```

pkg/controller/manifestwork/manifestwork_controller.go
```diff
@@ -77,6 +83,16 @@ func (r *ReconcileManifestWork) Reconcile(ctx context.Context, request reconcile
 		return reconcile.Result{}, err
 	}
 
+	mode, err := helpers.DetermineKlusterletMode(managedCluster)
+	if err != nil {
```
skeeey, line 87: same as auto import controller

pkg/helpers/helpers.go
```diff
@@ -528,3 +536,37 @@ func GetNodeSelector(cluster *clusterv1.ManagedCluster) (map[string]string, erro
 
 	return nodeSelector, nil
 }
+
+// DetermineKlusterletMode gets the klusterlet deploy mode for the managed cluster.
+func DetermineKlusterletMode(cluster *clusterv1.ManagedCluster) (string, error) {
```
skeeey, line 541: nit: how about only return the mode, use another method to handle `GetManagementCluster`

skeeey: for other cases, return `unknown`

cmd/manager/main.go
zhujian7: I tried, but it does not work. Do you know how to set it correctly?

pkg/controller/autoimport/autoimport_controller.go
zhujian7: Yep, I submitted another PR https://github.com/stolostron/managedcluster-import-controller/pull/124 implemented in the way you recommended. PTAL.

zhujian7:
https://github.com/stolostron/managedcluster-import-controller/pull/124 has been merged, close this.
/close

PR #124, opened 2022-02-15
Import cluster in hypershift detached mode

zhujian7:
This PR implements the same feature as https://github.com/stolostron/managedcluster-import-controller/pull/122, but in another way.
In this PR, we added a new `hypershift-detached-manifestwork` controller to process hypershift case, this will minimize the impact on the original processing logic.

Another additional change to the https://github.com/stolostron/managedcluster-import-controller/pull/122, we changed klusterlet deploy mode label to annotation

---
> description copied from https://github.com/stolostron/managedcluster-import-controller/pull/122

This PR supports importing a hypershift hosted cluster to the ACM hub in detached mode.
Users only need to create a managed cluster CR with
- annotation: `import.open-cluster-management.io/klusterlet-deploy-mode: Hypershift-Detached`
- annotation: `import.open-cluster-management.io/management-cluster-name: cluster1`
Then the import controller will start the importing process.

If users want to automatically import the hypershift hosted cluster, they are also required to provide an `auto-import-secret`. Or they can create the `external-managed-kubeconfig` on the management cluster manually.

And this feature can be controlled by a feature gate, it is disabled by default, you can enable it by adding a flag to the controller `--feature-gates=HypershiftImport=true`

Implement importing cluster in Detached mode according to the design https://github.com/stolostron/backlog/issues/18978#issuecomment-1021834313


pkg/controller/hypershiftdetached/manifests/external_managed_secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "external-managed-kubeconfig"
+  namespace: "{{ .KlusterletNamespace }}"
+type: Opaque
+data:
+  kubeconfig: {{ .ExternalManagedKubeconfig }}
```
skeeey, line 8: EOF

pkg/controller/importconfig/import_worker.go
```diff
@@ -0,0 +1,262 @@
...
+
+type importWorker interface {
+	generateImportSecret(ctx context.Context, managedCluster *clusterv1.ManagedCluster) (*corev1.Secret, error)
+}
+
+type workerFactory struct {
+	clientHolder *helpers.ClientHolder
+}
+
+func (f *workerFactory) newWorker(mode string) (importWorker, error) {
```
skeeey, line 30: cool! !
nit: could we separate this in different files and org them in a dir, like
```
importconfig
  |--works
         |-factory.go
         |-defualt_work.go
         |-detach_work.go
```

pkg/controller/manifestwork/manager.go
```diff
@@ -99,7 +100,12 @@ func add(importSecretInformer cache.SharedIndexInformer, mgr manager.Manager, r
 	if err := c.Watch(
 		&runtimesource.Kind{Type: &clusterv1.ManagedCluster{}},
 		&handler.EnqueueRequestForObject{},
-	); err != nil {
+		predicate.Predicate(predicate.Funcs{
+			GenericFunc: func(e event.GenericEvent) bool { return isDefaultModeObject(e.Object) },
```
skeeey, line 104: do we need this? I saw we seem already added the judgement in the controller

pkg/controller/importconfig/import_worker.go
zhujian7: Have put the workers in multiple files, but did move them to another folder since the global `log` variable is hard to process.

pkg/controller/manifestwork/manager.go
zhujian7: The judgment in the controller is in order to process requests from import secrets, functionally we can remove the predicate here, but it can filter the managed cluster earlier, so I added.

zhujian7:
@skeeey PTAL

skeeey:
I think this looks to me now, leave @elgnay to further review

/approve

pkg/controller/hypershiftdetached/controller.go
```diff
@@ -291,33 +258,83 @@ func createHypershiftDetachedManifestWork(managedCluster *clusterv1.ManagedClust
...
 		Spec: workv1.ManifestWorkSpec{
 			Workload: workv1.ManifestsTemplate{
 				Manifests: manifests,
 			},
 			DeleteOption: &workv1.DeleteOption{
-				PropagationPolicy: workv1.DeletePropagationPolicyTypeSelectivelyOrphan,
-				SelectivelyOrphan: &workv1.SelectivelyOrphan{
-					OrphaningRules: []workv1.OrphaningRule{
-						{
-							Resource:  "Secret",
```
zhujian7: replace `Secret` to `secrets`, I think this will work.

PR #137, opened 2022-03-09
wait for the addon manifestworks to be deleted

zhiweiyin318:
issue: https://github.com/stolostron/backlog/issues/20474

### DIFF
--- a/pkg/controller/manifestwork/manifestwork_controller.go
+++ b/pkg/controller/manifestwork/manifestwork_controller.go
@@ -142,18 +142,26 @@ func (r *ReconcileManifestWork) deleteManifestWorks(ctx context.Context, cluster
 	//
 	// if there are any Detached mode manifestworks we also wait for users to detach the managed cluster first.
 	ignoreKlusterletAndAddons := func(clusterName string, manifestWork workv1.ManifestWork) bool {
-		return manifestWork.GetName() == fmt.Sprintf("%s-%s", clusterName, klusterletSuffix) ||
-			manifestWork.GetName() == fmt.Sprintf("%s-%s", clusterName, klusterletCRDsSuffix) ||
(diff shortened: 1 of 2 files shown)


pkg/controller/manifestwork/manifestwork_controller.go
```diff
@@ -81,7 +83,17 @@ func (r *ReconcileManifestWork) Reconcile(ctx context.Context, request reconcile
 	}
 
 	if !managedCluster.DeletionTimestamp.IsZero() {
-		// the managed cluster is deleting, delete its manifestworks
+		// the managed cluster is deleting, delete its manifestWorks.
+		// all managedClusterAddons should be deleted before we delete the klusterlet ManifestWork since managedClusterAddons depends on ManifestWork.
+		addons := &addonv1alpha1.ManagedClusterAddOnList{}
```
skeeey, line 88: because we will stop the delete procession if there are addon manifestworks, my understanding only the addons is deleted, the manifestworks of addons will be deleted, so do we need this?

skeeey: or if we have this, could we remove the limitation for addon manifestworks? I think determine manifestworks may be  a better way?

pkg/controller/managedcluster/managedcluster_controller.go
```diff
@@ -220,6 +221,17 @@ func (r *ReconcileManagedCluster) deleteManagedClusterNamespace(
 		}
 	}
 
+	// managedClusterAddons depends on ManifestWork. so all managedClusterAddons should be deleted before we delete the ns.
+	addons := &addonv1alpha1.ManagedClusterAddOnList{}
```
skeeey, line 225: what's the case will make this happen?

skeeey:
We may need e2e to cover this

PR #141, opened 2022-03-14
avoid to add finalizer to manifestworks if the cluster is deleting

skeeey:


skeeey:
refer to https://github.com/stolostron/backlog/issues/20532

PR #155, opened 2022-04-13
[release 2.4]fix the length of sa name is over 63

zhiweiyin318:
fix issue: https://github.com/stolostron/backlog/issues/21634

### DIFF
--- a/pkg/controller/importconfig/cluster_info_test.go
+++ b/pkg/controller/importconfig/cluster_info_test.go
@@ -719,3 +719,31 @@ func TestGetImage(t *testing.T) {
 		t.Errorf("unexpected image: %v", newImage)
 	}
 }
+
+func TestGetBootstrapSAName(t *testing.T) {
+	cases := []struct {
+		name           string
+		clusterName    string
+		expectedSAName string
+		managedCluster *clusterv1.ManagedCluster
+	}{
+		{
+			name:           "short name",
+			clusterName:    "123456789",
+			expectedSAName: "123456789-bootstrap-sa",
+		},
+		{
+			name:           "long name",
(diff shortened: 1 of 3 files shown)


pkg/controller/importconfig/cluster_info_test.go
```diff
@@ -719,3 +719,31 @@ func TestGetImage(t *testing.T) {
...
+	}{
+		{
+			name:           "short name",
+			clusterName:    "123456789",
+			expectedSAName: "123456789-bootstrap-sa",
+		},
+		{
+			name:           "long name",
+			clusterName:    "123456789-123456789-123456789-123456789-123456789-123456789",
+			expectedSAName: "123456789-123456789-123456789-123456789-123456789--bootstrap-sa",
```
xuezhaojun, line 738: Are we sure here should be two `-`?

xuezhaojun: Maybe it's `123456789-123456789-123456789-123456789-123456789-bootstrap-sa` instead of `123456789-123456789-123456789-123456789-123456789--bootstrap-sa`?

xuezhaojun: Never mind, I got it. It will keep final length to 63.

xuezhaojun:
LGTM

pkg/controller/importconfig/cluster_info_test.go
zhiweiyin318: the rule is `<cluster-name>-bootstrap-sa`,I think it is ok to meet the rule if latest char is "-" in cluster name

PR #157, opened 2022-04-18
update imageRegistry api to fix issue 21706

zhiweiyin318:
fix issue: https://github.com/stolostron/backlog/issues/21706


go.mod
```diff
@@ -12,7 +12,6 @@ require (
 	github.com/openshift/hive/apis v0.0.0
 	github.com/openshift/library-go v0.0.0-20220112153822-ac82336bd076
 	github.com/spf13/pflag v1.0.5
-	github.com/stolostron/multicloud-operators-foundation v1.0.0-2021-10-26-20-16-14.0.20220110023249-172fb944faa9
```
zhujian7, line 15: since the multicloud-operators-foundation was deleted here, so do we still need the `replace` for `required by multicloud-operators-foundation`?

pkg/helpers/imageregistry/client.go
```diff
@@ -0,0 +1,150 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+)
+
+const (
+	// ClusterImageRegistryLabel value is namespace.managedClusterImageRegistry
+	ClusterImageRegistryLabel = "open-cluster-management.io/image-registry"
```
zhujian7, line 21: what the different between the ClusterImageRegistryLabel and ClusterImageRegistriesAnnotation?

pkg/helpers/imageregistry/client.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// ClusterImageRegistriesAnnotation value is a json string of ImageRegistries
+	ClusterImageRegistriesAnnotation = "open-cluster-management.io/image-registries"
+)
+
+type Registries struct {
```
zhujian7, line 27: Does this only represent only one registry or many registries?

go.mod
zhiweiyin318: I will open another PR to remove the lib, not sure which lib should not be upgraded in 2.5.

pkg/helpers/imageregistry/client.go
```diff
@@ -0,0 +1,150 @@
...
+	pullSecret := segs[1]
+	return c.kubeClient.CoreV1().Secrets(namespace).Get(context.TODO(), pullSecret, metav1.GetOptions{})
+}
+
+func (c *Client) ImageOverride(imageName string) string {
+	if len(c.imageRegistries.Registries) == 0 {
+		return imageName
+	}
+	overrideImageName := imageName
+	for i := 0; i < len(c.imageRegistries.Registries); i++ {
```
zhujian7, line 92: `for range` is simpler？

pkg/helpers/imageregistry/client.go
zhiweiyin318: ClusterImageRegistryLabel is to show which managedClusterImageRegistry is used in the mangedCluster,  ClusterImageRegistriesAnnotation is the value of image registries.

pkg/helpers/imageregistry/client.go
zhiweiyin318: only one registry. yeah it is confused. let me modify the name.

pkg/helpers/imageregistry/client.go
zhujian7: But I did not find any use of `ClusterImageRegistryLabel `.

pkg/helpers/imageregistry/client.go
zhiweiyin318: modify to Registry

pkg/helpers/imageregistry/client.go
zhiweiyin318: removed ClusterImageRegistryLabel which is not used anymore.

pkg/helpers/imageregistry/client.go
zhiweiyin318: just want to make it clean, the order can affect the results.

zhiweiyin318: let me add more doc here

zhiweiyin318: done

PR #159, opened 2022-04-20
fix nil cluster and add more ut

zhiweiyin318:

### DIFF
--- a/pkg/helpers/imageregistry/client.go
+++ b/pkg/helpers/imageregistry/client.go
@@ -56,13 +56,19 @@ func NewClient(kubeClient kubernetes.Interface) Interface {
 }
 
 func (c *Client) Cluster(cluster *clusterv1.ManagedCluster) Interface {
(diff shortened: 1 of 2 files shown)


pkg/helpers/imageregistry/client.go
```diff
@@ -56,13 +56,19 @@ func NewClient(kubeClient kubernetes.Interface) Interface {
 }
 
 func (c *Client) Cluster(cluster *clusterv1.ManagedCluster) Interface {
```
zhujian7, line 58: Should we return the error?

zhiweiyin318: no error returned because want to make it easy use `client.Cluster(cluster).PullSecret(),` `client.Cluster(cluster).ImageOverride(image)`.
if there error occurred here, will return error in  PullSecret(), and return original image in ImageOverride(image)

PR #161, opened 2022-04-22
Delete addons while detaching clusters

zhujian7:
Signed-off-by: zhujian <jiazhu@redhat.com>
issue reference: https://github.com/stolostron/backlog/issues/21835

### DIFF
--- a/pkg/controller/hosted/controller.go
+++ b/pkg/controller/hosted/controller.go
@@ -9,16 +9,17 @@ import (
 	"encoding/base64"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
 	"github.com/stolostron/managedcluster-import-controller/pkg/helpers"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorhelpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	workv1 "open-cluster-management.io/api/work/v1"
 
-	"github.com/openshift/library-go/pkg/operator/events"
-
 	"github.com/ghodss/yaml"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -26,7 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/kubernetes"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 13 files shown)


zhiweiyin318:
do we need `_debug_bin` file?

pkg/controller/manifestwork/manager.go
```diff
@@ -130,6 +131,26 @@ func add(importSecretInformer cache.SharedIndexInformer, mgr manager.Manager, r
 		return err
 	}
 
+	if err := c.Watch(
```
zhiweiyin318, line 134: I don't think we need watch addon. 
only need to delete all addon when the cluster is deleting.

zhujian7: Add this to trigger the reconcile to move forward at [here](https://github.com/stolostron/managedcluster-import-controller/pull/161/files#diff-768295ca0279c534580a99162d76fb98bc39e2b35f905dfaa34ca81d815b4574R174), otherwise, we need to wait the next resync period?

zhujian7:
@zhiweiyin318 PTAL and https://github.com/stolostron/backplane-operator/pull/181

pkg/controller/manifestwork/manager.go
zhiweiyin318: ok. looks good to me.

qiujian16:
I think we need import-controller to have permission to delete addons at first. Have we added that in installer?

pkg/controller/hosted/controller.go
```diff
@@ -102,6 +102,9 @@ func (r *ReconcileHosted) Reconcile(ctx context.Context, request reconcile.Reque
 	}
 
 	if !managedCluster.DeletionTimestamp.IsZero() {
+		if err := helpers.DeleteManagedClusterAddons(ctx, r.clientHolder.RuntimeClient, managedClusterName); err != nil {
```
qiujian16, line 105: I think we should still go head deleting works even we have error with addon deletion. Maybe we should delete work and addon in one single func and return aggregated error.

pkg/controller/hosted/controller.go
```diff
@@ -238,16 +241,33 @@ func (r *ReconcileHosted) deleteManifestWorks(ctx context.Context, cluster *clus
...
+		return true
 	}
 	err := helpers.DeleteManifestWorkWithSelector(ctx, r.clientHolder.RuntimeClient, r.recorder, cluster, works, ignoreAddons)
 	if err != nil {
 		return err
 	}
 
+	noAddons, err := helpers.NoManagedClusterAddons(ctx, r.clientHolder.RuntimeClient, cluster.GetName())
+	if err != nil {
+		return err
```
qiujian16, line 264: I have not take a deep look at this code before, but we should consider requeue a certain time instead of error fallback if there are still returning addon or works. We might not need to do this in this PR, but we should consider it later 

cc @zhiweiyin318

pkg/controller/hosted/controller.go
zhiweiyin318: > I think we need import-controller to have permission to delete addons at first. Have we added that in installer?

@qiujian16 yes, delete rbac has been added into installer.

zhiweiyin318:
> I think we need import-controller to have permission to delete addons at first. Have we added that in installer?

@qiujian16 yes, delete rbac has been added into installer.

pkg/controller/hosted/controller.go
```diff
@@ -221,44 +221,85 @@ func (r *ReconcileHosted) getHostedManifestWork(ctx context.Context, namespace,
...
+		return reconcile.Result{}, err
 	}
 	if !noPending {
 		// still have other works, do nothing
-		return nil
+		return reconcile.Result{}, nil
 	}
 
-	return r.deleteHostedManifestWorks(ctx, r.clientHolder.RuntimeClient, r.recorder, cluster)
+	return reconcile.Result{}, r.deleteHostedManifestWorks(ctx, r.clientHolder.RuntimeClient, r.recorder, cluster)
```
qiujian16, line 302: should we also RequeueAfter here?

zhujian7: I don't think it is necessary, if `deleteHostedManifestWorks` returns err, it will requeue. And the `deleteHostedManifestWorks` func only return get/delete manifestwork error which should be retried immediately

qiujian16: what i mean is when you have pending works, should you return error which requeu immediately, or should it requeue after a certain time.

zhujian7: if the work is pending, it will return nil. see [here](https://github.com/stolostron/managedcluster-import-controller/blob/cc6cab558ff07ea844bb0c106f53350f900b352e/pkg/helpers/manifestwork.go#L127)

qiujian16: hrm, I think we should discuss about that, but it is not a blocker for this PR.

PR #171, opened 2022-06-28
support customizing klusterlet ns with annotation

elgnay:
Related issue: https://github.com/stolostron/backlog/issues/23554
Signed-off-by: Yang Le <yangle@redhat.com>

### DIFF
--- a/pkg/controller/importconfig/importconfig_controller.go
+++ b/pkg/controller/importconfig/importconfig_controller.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 
+	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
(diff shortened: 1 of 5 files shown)


pkg/controller/importconfig/importconfig_controller.go
```diff
@@ -136,3 +137,11 @@ func (r *ReconcileImportConfig) Reconcile(ctx context.Context, request reconcile
 
 	return reconcile.Result{}, nil
 }
+
+func klusterletNamespace(managedCluster *clusterv1.ManagedCluster) string {
+	if klusterletNamespace, ok := managedCluster.Annotations[constants.KlusterletNamespaceAnnotation]; ok {
+		return klusterletNamespace
```
skeeey, line 143: Do we need to check `The namespace must have a prefix of "open-cluster-management-"`

zhujian7:
LGTM

pkg/controller/importconfig/importconfig_controller.go
skeeey: This will be validated in klusterlet api

PR #172, opened 2022-06-30
Improve coverage score

xuezhaojun:
Signed-off-by: xuezhaojun <zxue@redhat.com>

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,7 +1,7 @@
 sonar.projectKey=open-cluster-management_managedcluster-import-controller
 sonar.projectName=managedcluster-import-controller
 sonar.sources=.
-sonar.exclusions=cmd/manager/*.go,pkg/controller/*.go,pkg/controller/**/manager.go,**/*_test.go,test/**
+sonar.exclusions=cmd/manager/*.go,pkg/controller/*.go,pkg/controller/**/manager.go,**/*_test.go,test/**,**/manifestwork.go
 sonar.tests=.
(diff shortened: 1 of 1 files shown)


sonar-project.properties
```diff
@@ -1,7 +1,7 @@
 sonar.projectKey=open-cluster-management_managedcluster-import-controller
 sonar.projectName=managedcluster-import-controller
 sonar.sources=.
-sonar.exclusions=cmd/manager/*.go,pkg/controller/*.go,pkg/controller/**/manager.go,**/*_test.go,test/**
+sonar.exclusions=cmd/manager/*.go,pkg/controller/*.go,pkg/controller/**/manager.go,**/*_test.go,test/**,**/manifestwork.go
```
xuezhaojun, line 4: @zhujian7  I suppose `manifestwork.go` test is covered by e2e, is that right?

zhujian7: I think the cases are covered in e2e. cc @skeeey

skeeey: yes, but I think we may add need some unit tests for this if we have time :)

and for import controller e2e, we generated the coverage, and the coverage is 

```
TOTAL COVERAGE IS 74.1%
```

you can find it from our e2e test

I think the problem seems that we cannot add this coverage to sonar, because the e2e run with the github action

xuezhaojun: Cool, I guess if we have covered this part in e2e, we can exclude `manifestwork.go` for now and add u-t later.

PR #173, opened 2022-07-01
support kube 1.24 sa secret changes

skeeey:

### DIFF
--- a/pkg/controller/importconfig/cluster_info.go
+++ b/pkg/controller/importconfig/cluster_info.go
@@ -32,55 +32,54 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// getBootstrapSecret looks for the bootstrap secret from bootstrap sa
+// getBootstrapSecret lists the secrets from the managed cluster namespace to look
+// for the managed cluster bootstrap token secret.
+//
+// Note: this function depends on the OCP service account feature.
+// Starting with kube 1.24 (ocp 4.11), the k8s won't generate secrets any longer
+// automatically for ServiceAccounts, for OCP, when a service account is created,
+// the OCP will create two secrets, one stores dockercfg with name format (<sa name>-dockercfg-<random>)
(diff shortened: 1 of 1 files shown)


skeeey:
refer to https://github.com/stolostron/backlog/issues/23917

pkg/controller/importconfig/cluster_info.go
```diff
@@ -317,3 +323,12 @@ func getBootstrapSAName(clusterName string) string {
 	}
 	return bootstrapSAName
 }
+
+func getBootstrapSASecretName(clusterName string) string {
+	bootstrapSASecretName := fmt.Sprintf("%s-%s", clusterName, bootstrapSASecretSuffix)
+	if len(bootstrapSASecretName) > 63 {
+		return fmt.Sprintf("%s-%s", clusterName[:63-len("-"+bootstrapSASecretSuffix)], bootstrapSASecretSuffix)
```
qiujian16, line 330: this needs doc. Why secret name is this?

pkg/controller/importconfig/cluster_info.go
```diff
@@ -34,12 +34,15 @@ import (
 
 // getBootstrapSecret looks for the bootstrap secret from bootstrap sa
 func getBootstrapSecret(ctx context.Context, kubeClient kubernetes.Interface, managedCluster *clusterv1.ManagedCluster) (*corev1.Secret, error) {
+	secretNames := []string{getBootstrapSASecretName(managedCluster.Name)}
+
```
qiujian16, line 38: this needs more doc, because it depends on a ocp specific function.

pkg/controller/importconfig/cluster_info.go
skeeey: remove this, I use the service account name as the secret name

pkg/controller/importconfig/cluster_info.go
skeeey: add more doc, and this dose not depend on the ocp specific function, it only is related to kube version

PR #174, opened 2022-07-01
avoid delete klusterlet on the managed cluster

skeeey:

### DIFF
--- a/pkg/controller/autoimport/autoimport_controller.go
+++ b/pkg/controller/autoimport/autoimport_controller.go
@@ -7,14 +7,17 @@ import (
 	"context"
 	"fmt"
 
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	workv1 "open-cluster-management.io/api/work/v1"
+
 	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
 	"github.com/stolostron/managedcluster-import-controller/pkg/helpers"
-	clusterv1 "open-cluster-management.io/api/cluster/v1"
 
 	"github.com/openshift/library-go/pkg/operator/events"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 
@@ -55,7 +58,11 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	// TODO: we will use list instead of get to reduce the request in the future
+	if helpers.DetermineKlusterletMode(managedCluster) != constants.KlusterletDeployModeDefault {
+		return reconcile.Result{}, nil
+	}
+
+	// TODO: we will use lister instead of get to reduce the request in the future
 	autoImportSecret, err := r.kubeClient.CoreV1().Secrets(managedClusterName).Get(ctx, constants.AutoImportSecretName, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
 		// the auto import secret could have been deleted, do nothing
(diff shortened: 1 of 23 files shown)


skeeey:
refer to https://github.com/stolostron/backlog/issues/24017

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -79,6 +81,16 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
 	}
 
+	// ensure the klusterlet manifest works exist
+	listOpts := &client.ListOptions{Namespace: managedClusterName}
+	manifestWorks := &workv1.ManifestWorkList{}
+	if err := r.client.List(ctx, manifestWorks, listOpts); err != nil {
+		return reconcile.Result{}, err
+	}
+	if !helpers.HasKlusterletManifestWorks(managedClusterName, manifestWorks) {
+		return reconcile.Result{}, fmt.Errorf("there are no klusterlet manifest works for managed cluster %s", managedClusterName)
```
qiujian16, line 91: should we return error or nil? I think if we watch the work here, we can return nil here.

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -79,6 +81,16 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	// ensure the klusterlet manifest works exist
+	listOpts := &client.ListOptions{Namespace: managedClusterName}
+	manifestWorks := &workv1.ManifestWorkList{}
+	if err := r.client.List(ctx, manifestWorks, listOpts); err != nil {
```
qiujian16, line 87: do you think we should have a labelSelector here?

pkg/controller/manifestwork/manifestwork_controller.go
```diff
@@ -208,34 +205,31 @@ func (r *ReconcileManifestWork) deleteManifestWorks(
...
 
 	return reconcile.Result{}, helpers.DeleteManifestWork(
 		ctx, r.clientHolder.RuntimeClient, r.recorder, klusterletWork.Namespace, klusterletWork.Name)
 }
 
 func createKlusterletCRDsManifestWork(managedCluster *clusterv1.ManagedCluster, importSecret *corev1.Secret) *workv1.ManifestWork {
 	crdsKey := constants.ImportSecretCRDSV1YamlKey
-	if !helpers.IsAPIExtensionV1Supported(managedCluster.Status.Version.Kubernetes) {
-		log.Info("crd v1 is not supported, add v1beta1")
+	if managedCluster.Status.Version.Kubernetes != "" &&
```
qiujian16, line 230: managedCluster.Status.Version.Kubernetes != "" this seems redundant.

pkg/controller/autoimport/autoimport_controller.go
skeeey: I don't want to watch the work here :), the work is created along with the import-secret, we already watch the import-secret, in most conditions, this error will not happen, and we use the controller-runtime, it's hard to customize the work informer, so if we watch the work, I think it will bring extra cost

pkg/controller/autoimport/autoimport_controller.go
skeeey: yeah, we can have a label selector

pkg/controller/manifestwork/manifestwork_controller.go
skeeey: it's not redundant, because if the version is empty, the  IsAPIExtensionV1Supported will return false

pkg/controller/autoimport/autoimport_controller.go
skeeey: use watching the manifest works instead of watching the import-secret

pkg/controller/autoimport/manager.go
```diff
@@ -46,20 +53,37 @@ func add(importSecretInformer, autoImportSecretInformer cache.SharedIndexInforme
 		return err
 	}
 
-	// watch the import secrets
+	// watch the klusterlet manifest works
 	if err := c.Watch(
-		source.NewImportSecretSource(importSecretInformer),
-		&source.ManagedClusterSecretEventHandler{},
+		&runtimesource.Kind{Type: &workv1.ManifestWork{}},
```
qiujian16, line 58: I think you should watch import secret and work both. The key could be namespace.

skeeey: the key is namespace (`managedClusterName := request.Namespace`) , so we don't need a special handling

ok, add the import secret back, it will help to resolve corner case

PR #179, opened 2022-07-13
add ns deletion controller

qiujian16:
Signed-off-by: Jian Qiu <jqiu@redhat.com>

### DIFF
--- a/pkg/controller/clusternamespacedeletion/controller.go
+++ b/pkg/controller/clusternamespacedeletion/controller.go
@@ -1,0 +1,162 @@
+// Copyright (c) Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package clusternamespacedeletion
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+
+	asv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
+	clustercontroller "github.com/stolostron/managedcluster-import-controller/pkg/controller/managedcluster"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	addonv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var log = logf.Log.WithName(controllerName)
+
+const (
+	curatorJobPrefix  string = "curator-job"
+	postHookJobPrefix string = "posthookjob"
+	preHookJobPrefix  string = "prehookjob"
+
+	deletionGracePeriod = 10 * time.Minute
+)
+
+// ReconcileClusterNamespaceDeletion delete cluster namespace when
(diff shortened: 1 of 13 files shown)


pkg/controller/clusternamespacedeletion/controller.go
```diff
@@ -0,0 +1,161 @@
...
+	client   client.Client
+	recorder events.Recorder
+}
+
+// blank assignment to verify that ReconcileManagedCluster implements reconcile.Reconciler
+var _ reconcile.Reconciler = &ReconcileClusterNamespaceDeletion{}
+
+func (r *ReconcileClusterNamespaceDeletion) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Name", request.Name)
+	reqLogger.Info("Reconciling the managed cluster meta object")
```
skeeey, line 53: Reconciling the managed cluster namespace deletion?

skeeey, line 53: Reconciling the managed cluster namespace deletion?

pkg/controller/clusternamespacedeletion/controller.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := r.client.List(ctx, pods, client.InNamespace(ns.Name)); err != nil {
+		return reconcile.Result{}, err
+	}
+	validPods := filterPods(pods.Items, ns.Name)
+	if len(validPods) > 0 {
+		klog.Warningf("ManagedClusterNamespaceRemained", "There are some pods remaining in namespace %s", ns.Name)
+		return reconcile.Result{RequeueAfter: deletionGracePeriod}, nil
+	}
+
+	err = r.client.Delete(ctx, ns)
```
skeeey, line 136: may need an event/log here?

pkg/controller/clusternamespacedeletion/manager.go
```diff
@@ -0,0 +1,128 @@
...
+		&source.Kind{Type: &clusterv1.ManagedCluster{}},
+		handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
+			return []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Name: o.GetName(),
+					},
+				},
+			}
+		}),
```
skeeey, line 62: do we need a predicate here? only delete/update?

pkg/controller/clusternamespacedeletion/manager.go
```diff
@@ -0,0 +1,128 @@
...
+	); err != nil {
+		return err
+	}
+
+	if err := c.Watch(
+		&source.Kind{Type: &corev1.Namespace{}},
```
skeeey, line 68: It seems we don't need to watch the ns,

pkg/controller/clusternamespacedeletion/manager.go
```diff
@@ -0,0 +1,128 @@
...
+				},
+			}
+		}),
+	); err != nil {
+		return err
+	}
+
+	if err := c.Watch(
+		&source.Kind{Type: &hivev1.ClusterDeployment{}},
+		handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
```
skeeey, line 84: do we need a predicate here? only delete/update?

pkg/controller/managedcluster/managedcluster_controller.go
```diff
@@ -120,7 +110,7 @@ func (r *ReconcileManagedCluster) Reconcile(ctx context.Context, request reconci
 	}
 
 	// managed cluster is deleting, remove its namespace
```
skeeey, line 112: the comments is not right

pkg/controller/clusternamespacedeletion/manager.go
qiujian16: I think we need to in case controller restarted

skeeey:
It's better to have e2e test for `infraenv` and `curator`, I think we can add the `infraenv` part firstly

PR #206, opened 2022-12-23
using hosted cluster name label filter the hosted works

skeeey:


pkg/controller/hosted/controller.go
```diff
@@ -300,7 +300,8 @@ func createHostedManifestWork(managedClusterName string,
 			Name:      hostedKlusterletManifestWorkName(managedClusterName),
 			Namespace: manifestWorkNamespace,
 			Labels: map[string]string{
-				constants.HostedWorksLabel: "true",
+				constants.HostedWorksLabel:   "true",
```
zhujian7, line 303: do we still need this label?

skeeey: We need this, we use label to informer hosted works https://github.com/stolostron/managedcluster-import-controller/blob/main/cmd/manager/main.go#L171 to reduce the cache

skeeey: We can only check the hosted cluster label, will remove this

PR #208, opened 2023-02-24
Add condition for cluster auto importing

zhujian7:


pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -78,100 +77,233 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
+				newManagedClusterImportSucceededCondition(
+					metav1.ConditionFalse,
+					constants.ConditionReasonAutoImportSecretInvalid,
+					fmt.Sprintf("Retry to import managed cluster, retry times: %d/%d, error: %v",
+						rc.current, rc.total, err),
+				), rc, err
 		}
 
-		// failed to apply the import secrect, reduce the retry times and reconcile again
-		retryErr := helpers.UpdateAutoImportRetryTimes(ctx, r.kubeClient, r.recorder, autoImportSecret.DeepCopy())
```
skeeey, line 134: how about to enhance the `UpdateAutoImportRetryTimes`? because this is also is used by hosted import controller, I think we should need a  consistent way to handle the auto import retry.

For `UpdateAutoImportRetryTimes`, we may consider to add a currentRetryTimes in the autoimport secret data instead of reducing the `autoImportRetry` directly

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -78,100 +77,233 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
+	autoImportSecret *v1.Secret) (reconcile.Result, *metav1.Condition, importRetryCount, error) {
+
+	reqLogger := log.WithValues("Request.Namespace", managedClusterName)
+
+	var err error
+	rc := importRetryCount{
+		total:   1,
+		current: 0,
+	}
+	if len(autoImportSecret.Data[constants.AutoImportRetryName]) != 0 {
```
skeeey, line 164: this is a good to validate the autoimport secret format firstly, we may have a validate function for this part to make sure the autoimport secret format is correct (also consider the hosted import)

zhujian7:
/close
duplicate with https://github.com/stolostron/managedcluster-import-controller/pull/209

PR #209, opened 2023-03-01
Add condition for cluster importing

zhujian7:
1. Extract the common logic of import into helpers.ImportHelper
2. Add a new importstatus controller to check whether the klusterlet manifestwork is available
3. Add some condition details for importing clusters

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -100,7 +100,8 @@ const (
 const (
 	// ConditionManagedClusterImportSucceeded is the condition type of managed cluster to indicate whether the managed
 	// cluster is imported successfully
-	ConditionManagedClusterImportSucceeded        = "ManagedClusterImportSucceeded"
+	ConditionManagedClusterImportSucceeded = "ManagedClusterImportSucceeded"
+
 	ConditionReasonAutoImportSecretInvalid        = "AutoImportSecretInvalid"
 	ConditionReasonImportSecretNotReady           = "ImportSecretNotFound"
 	ConditionReasonKlusterletManifestWorkNotReady = "KlusterletManifestWorkNotFound"
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -14,9 +14,13 @@ const (
 	// AutoImportRetryName is the secret data key of auto import retry
 	AutoImportRetryName string = "autoImportRetry"
 
-	// AnnotationAutoImportCurrentRetry is the annotation key of auto import secret used for indicate
+	// AnnotationAutoImportCurrentRetry is the annotation key of auto import secret used to indicate
 	// the current retry times of auto importing a managed cluster
-	AnnotationAutoImportCurrentRetry = "import.open-cluster-management.io/current-retry"
+	AnnotationAutoImportCurrentRetry = "managedcluster-import-controller.open-cluster-management.io/current-retry"
+
+	// AnnotationKeepingAutoImportSecret is the annotation key of auto import secret used to indicate
+	// keeping this secret after the cluster is imported successfully
+	AnnotationKeepingAutoImportSecret = "managedcluster-import-controller.open-cluster-management.io/keeping-auto-import-secret"
 )
 
 const PodNamespaceEnvVarName = "POD_NAMESPACE"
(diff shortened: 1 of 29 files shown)


pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -69,109 +77,113 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
-		// the auto import secret could have been deleted, do nothing
+	if !managedCluster.DeletionTimestamp.IsZero() {
 		return reconcile.Result{}, nil
 	}
-	if err != nil {
-		return reconcile.Result{}, err
-	}
 
-	reqLogger.Info("Reconciling auto import secret")
+	if helpers.ClusterNotNeedToApplyImportResources(managedCluster) {
```
zhujian7, line 84: Or we can separate the `ManagedClusterImportSucceeded` into two condition `ImportingResourcesApplied`(autoimport/selfmanagement/clusterdeployment) and `ManagedClusterImportSucceeded`(importstatus) ？

pkg/constants/constants.go
```diff
@@ -90,3 +103,17 @@ const (
 	KlusterletSuffix     = "klusterlet"
 	KlusterletCRDsSuffix = "klusterlet-crds"
 )
+
+const (
+	// ConditionManagedClusterImportSucceeded is the condition type of managed cluster to indicate whether the managed
+	// cluster is imported successfully
```
zhiweiyin318, line 109: should we consider to move the condition definitions  to the [clusterlifecycle-api](https://github.com/stolostron/cluster-lifecycle-api) ? if there are other components depend on these conditions?

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -78,100 +90,86 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-	if len(manifestWorks) != 2 {
-		reqLogger.Info(fmt.Sprintf("Waiting for klusterlet manifest works for managed cluster %s", managedClusterName))
-		return reconcile.Result{}, nil
+	if len(autoImportSecret.Data[constants.AutoImportRetryName]) != 0 {
+		totalRetry, err = strconv.Atoi(string(autoImportSecret.Data[constants.AutoImportRetryName]))
+		if err != nil {
+			r.recorder.Warningf("AutoImportRetryInvalid",
```
skeeey, line 107: nit, we may not need this warning, because we already have the condition

pkg/helpers/autoimport.go
```diff
@@ -6,58 +6,248 @@ package helpers
...
+}
+
+const (
+	conditionMessageImportingResourcesApplied = "Importing resources are applied, wait for resources be available"
+)
+
+func ImportingResourcesApplied(condition *metav1.Condition) bool {
+	if condition != nil && condition.Type == constants.ConditionManagedClusterImportSucceeded &&
+		condition.Reason == constants.ConditionReasonManagedClusterImporting &&
+		condition.Message == conditionMessageImportingResourcesApplied {
```
skeeey, line 249: do we need judge message?

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -78,100 +90,86 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
+	totalRetry := 1
+	if len(autoImportSecret.Annotations[constants.AnnotationAutoImportCurrentRetry]) != 0 {
+		lastRetry, err = strconv.Atoi(autoImportSecret.Annotations[constants.AnnotationAutoImportCurrentRetry])
+		if err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
-	// ensure the klusterlet manifest works exist
-	workSelector := labels.SelectorFromSet(map[string]string{constants.KlusterletWorksLabel: "true"})
```
skeeey, line 94: I think we need keep this check here to avoid apply the manifestwork early for backup-restore case

pkg/helpers/autoimport.go
zhujian7: yes, because there are other cases the reason is `importing`,  so I think we can add a specific method to set the the ImportSucceeded condition to ResourceApplied

pkg/constants/constants.go
zhujian7: ok, will submit a follow up PR to do this.

PR #211, opened 2023-03-21
filter addon manifestwork with label

zhiweiyin318:
fix issue : https://issues.redhat.com/browse/ACM-4232


pkg/controller/manifestwork/manifestwork_controller.go
```diff
@@ -169,6 +170,16 @@ func (r *ReconcileManifestWork) deleteManifestWorks(
 	//
 	// if there are any Hosted mode manifestworks we also wait for users to detach the managed cluster first.
 	ignoreKlusterletAndAddons := func(clusterName string, manifestWork workv1.ManifestWork) bool {
+		// if the addon is implemented by addon-framekwork, the format of manifestwork name is
+		// the old format : addon-<addon name>-deploy,
+		// the new format : addon-<addon name>-deploy-<index>
+		// and the manifestwork has the addon label.
+		// so cannot only use the name to filter.
+		workLabels := manifestWork.GetLabels()
+		if _, ok := workLabels[addonapiv1alpha1.AddonLabelKey]; ok {
```
skeeey, line 179: do we need to consider hosted part?

zhiweiyin318: added

skeeey:
LGTM

/cc @zhujian7

pkg/controller/hosted/controller.go
```diff
@@ -351,6 +352,15 @@ func (r *ReconcileHosted) deleteManifestWorks(ctx context.Context, cluster *clus
 	// delete works that do not include klusterlet works and klusterlet addon works, the addon works were removed
 	// above, we need to wait them to be deleted.
 	ignoreAddons := func(clusterName string, manifestWork workv1.ManifestWork) bool {
+		// if the addon is implemented by addon-framekwork, the format of manifestwork name is
+		// the old format : addon-<addon name>-deploy,
+		// the new format : addon-<addon name>-deploy-<index>
+		// and the manifestwork has the addon label.
+		// so cannot only use the name to filter.
+		workLabels := manifestWork.GetLabels()
+		if _, ok := workLabels[addonapiv1alpha1.AddonLabelKey]; ok {
```
zhujian7, line 361: for hosted mode addon, I think we need to check the addon namespace label

zhiweiyin318: here only delete the manifestworks in hosted cluster ns not hosting cluster ns, so addon name label is ok, right?

PR #215, opened 2023-03-22
validate hosting cluster

zhujian7:
issue reference: https://issues.redhat.com/browse/ACM-4255

### DIFF
--- a/pkg/controller/hosted/controller.go
+++ b/pkg/controller/hosted/controller.go
@@ -81,9 +81,6 @@ func (r *ReconcileHosted) Reconcile(ctx context.Context, request reconcile.Reque
 		return reconcile.Result{}, nil
 	}
 
-	// TODO(zhujian7): check if annotation hosting cluster is provided, check if the hosting cluster
-	// is a managed cluster of hub, and check its status.
-
 	reqLogger.Info("Reconciling the manifest works of the hosted mode managed cluster")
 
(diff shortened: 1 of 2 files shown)


pkg/controller/hosted/controller.go
```diff
@@ -174,7 +171,25 @@ func (r *ReconcileHosted) importCluster(ctx context.Context,
...
 	}
 
-	hostedWorks, err := r.informerHolder.HostedWorkLister.ManifestWorks(hostingCluster).List(hostedWorksSelector)
+	hostingCluster := &clusterv1.ManagedCluster{}
+	err = r.clientHolder.RuntimeClient.Get(ctx, types.NamespacedName{Name: hostingClusterName}, hostingCluster)
+	if err != nil {
+		return reconcile.Result{},
+			helpers.NewManagedClusterImportSucceededCondition(metav1.ConditionFalse,
+				constants.ConditionReasonManagedClusterImporting,
+				fmt.Sprintf("Validate if the hosting cluster is a managed cluster of the hub, error: %v", err)),
```
skeeey, line 180: nit: this message may be confused, how about use failed to get the managed cluster %s

skeeey: And we may only need to focus the notfound error,  for other errors, when the errors happed, we will update condition, so the same error may be also happed for update action

PR #221, opened 2023-05-06
refresh bootstrap kubeconfig once kube apiserver and CA changes

elgnay:
The related issue: https://issues.redhat.com/browse/ACM-4830

### DIFF
--- a/pkg/controller/importconfig/cluster_info.go
+++ b/pkg/controller/importconfig/cluster_info.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/importconfig/cluster_info.go
```diff
@@ -259,9 +261,26 @@ func getBootstrapKubeConfigData(ctx context.Context, clientHolder *helpers.Clien
 	}
 
 	kubeConifgData := getBootstrapKubeConfigDataFromImportSecret(importSecret)
+	kubeAPIServer, caData, token, err := parseKubeConfigData(kubeConifgData)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// check if the kube apiserver address is changed
+	validKubeAPIServer, err := validateKubeAPIServerAddress(ctx, kubeAPIServer, clientHolder)
```
zhujian7, line 270: all these checking also need to get infrastructure and apiserver, why do not we just `return createKubeConfig(ctx, clientHolder, cluster)` directly?

elgnay: `createKubeConfig(ctx, clientHolder, cluster)` will create a new token. If the current bootstrap kubeconfig is valid, it is not necessary to create a new one, which will cause the restarting of the klusterlet agents.

PR #224, opened 2023-05-11
ignore copyright check for tekton

skeeey:


skeeey:
force merged, ignore the Red Hat Trusted App Pipeline check

PR #249, opened 2023-07-12
Replace `wait.Poll` to `Eventually`.

xuezhaojun:
## Motivation: 

`wait.Poll` can't tell us the e2e fail on which line exactly. No matter what reason cause the timeout, we can only get:
```
<*errors.errorString | 0xc0001c59c0>: {
          s: "timed out waiting for the condition",
      }
      timed out waiting for the condition
```

### DIFF
--- a/test/e2e/clusterdeployment_test.go
+++ b/test/e2e/clusterdeployment_test.go
@@ -14,7 +14,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/rand"
-	"k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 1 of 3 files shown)


test/e2e/clusterdeployment_test.go
```diff
@@ -127,54 +130,60 @@ func assertKlusterletNamespaceDeleted() {
...
+		}, 10*time.Minute, 1*time.Second).Should(gomega.Succeed())
 	})
 
 }
 
 func assertKlusterletDeleted() {
 	start := time.Now()
 	ginkgo.By("Should delete the klusterlet crd", func() {
-		gomega.Expect(wait.Poll(1*time.Second, 10*time.Minute, func() (bool, error) {
+		gomega.Eventually(func() error {
```
xuezhaojun, line 151: For example, no matter what reason cause timeout, we will always get the error:
```
<*errors.errorString | 0xc0001c59c0>: {
          s: "timed out waiting for the condition",
      }
      timed out waiting for the condition
```

PR #255, opened 2023-07-19
Sperate e2e scripts.

xuezhaojun:


.github/workflows/e2e-tests.yml
```diff
@@ -19,6 +19,12 @@ jobs:
       with:
         go-version: 1.19
 
+    - name: Set up kubectl
```
xuezhaojun, line 22: Use git action to provide kubectl to save some code.

Makefile
```diff
@@ -86,20 +86,19 @@ clean: clean-e2e-test
...
-## Build e2e test binary
-.PHONY: build-e2e-test
-build-e2e-test:
-	go test -c ./test/e2e -o _output/e2e.test
-
 ## Runs e2e test
 .PHONY: e2e-test
-e2e-test: build-image build-e2e-test
-	@build/run-e2e-tests.sh
+e2e-test: build-image
```
xuezhaojun, line 91: By separate script, we can reuse `setup-ocm.sh` and `setup-import-controller.sh` when we want to test against OCP env.

Makefile
```diff
@@ -34,7 +34,7 @@ export DOCKER_FILE        = $(BUILD_DIR)/Dockerfile
 export DOCKER_IMAGE      ?= $(COMPONENT_NAME)
 export DOCKER_BUILDER    ?= docker
 
-BEFORE_SCRIPT := $(shell build/before-make.sh)
+KUBECONFIG?=./.kubeconfig
```
xuezhaojun, line 37: Move kubeconfig under the project root instead of the _output/ because we want to run:

```
go test -c ./test/e2e -o _output/e2e.test
_output/e2e.test -test.v -ginkgo.v
```

At the same layer.

PR #258, opened 2023-07-24
Fix prow e2e.

xuezhaojun:


test/e2e/managedcluter_test.go
```diff
@@ -20,7 +20,8 @@ import (
 	"github.com/stolostron/managedcluster-import-controller/test/e2e/util"
 )
 
-var _ = ginkgo.Describe("Importing a managed cluster manually", func() {
+// ginkgo.Serial promise specs not run in parallel
+var _ = ginkgo.Describe("Importing a managed cluster manually", ginkgo.Serial, func() {
```
xuezhaojun, line 24: Both 3 cases:
* https://github.com/stolostron/managedcluster-import-controller/blob/9175cd33621f745f21fca39d5b101e7a880fb60d/test/e2e/managedcluter_test.go#L57
* https://github.com/stolostron/managedcluster-import-controller/blob/9175cd33621f745f21fca39d5b101e7a880fb60d/test/e2e/managedcluter_test.go#L75
* https://github.com/stolostron/managedcluster-import-controller/blob/9175cd33621f745f21fca39d5b101e7a880fb60d/test/e2e/managedcluter_test.go#L153

So it should run in serial to avoid conflict.

PR #267, opened 2023-08-10
add update/patch permission of namespace to klusterlet for 2.3

elgnay:


xuezhaojun:
This one is already fixed by a cherrypick PR: https://github.com/stolostron/managedcluster-import-controller/pull/298
@elgnay Should we close this?

PR #274, opened 2023-08-15
bump golangci-lint to 1.54.1

zhujian7:
Fix oom issue https://github.com/golangci/golangci-lint/issues/3538


zhujian7:
cc @zhiweiyin318 @elgnay

PR #278, opened 2023-08-17
Start agent with singleton mode

qiujian16:


pkg/controller/importconfig/manifests/klusterlet/klusterlet.yaml
```diff
@@ -11,11 +11,12 @@ spec:
 {{- if eq .InstallMode "Hosted"}}
     mode: Hosted
 {{- else }}
-    mode: Default
+    mode: Singleton
 {{- end}}
   registrationImagePullSpec: "{{ .RegistrationImageName }}"
   workImagePullSpec: "{{ .WorkImageName }}"
   clusterName: "{{ .ManagedClusterNamespace }}"
+  imagePullSpec: "{{ .ImageName }}"
```
xuezhaojun, line 19: How about using `imagePullSepc: "{{ . RegistrationOperatorImage}}"`, it's the same value.

pkg/controller/importconfig/worker_default.go
```diff
@@ -71,6 +71,7 @@ func (w *defaultWorker) generateImportSecret(ctx context.Context, managedCluster
 			BootstrapKubeconfig:     base64.StdEncoding.EncodeToString(bootstrapKubeconfigData),
 			RegistrationImageName:   registrationImageName,
 			WorkImageName:           workImageName,
+			ImageName:               registrationOperatorImageName,
```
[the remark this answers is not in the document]
qiujian16: hrm it is the same value right now, since we use operator binary to run singleton agent. It is possible in the future that we run a more customized binary rather than using operator. I think we should avoid the confusion here.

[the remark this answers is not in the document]
xuezhaojun: Got it.

pkg/controller/importconfig/importconfig_controller.go
```diff
@@ -91,7 +91,7 @@ func (r *ReconcileImportConfig) Reconcile(ctx context.Context, request reconcile
 	switch mode {
 	case constants.KlusterletDeployModeDefault:
 		yamlcontent, err = bootstrap.NewKlusterletManifestsConfig(
-			operatorv1.InstallModeDefault,
```
xuezhaojun, line 94: Just to confirm, the install mode origin is from this line of code: https://github.com/stolostron/managedcluster-import-controller/blob/94cefd12dc428f999991dfeabc75fdac6b75ba2f/pkg/helpers/helpers.go#L642

`mode, ok := cluster.Annotations[constants.KlusterletDeployModeAnnotation]` From this annotation, we still only have 2 modes right? 

Only now we render with `Mode:Singleton` when we extract `Default` from the annotation?

qiujian16: yes

qiujian16: let me think about it, otherwise we need ui to update

qiujian16: updated, if annotation is not set, use the singleton by default, otherwise read the annotation accordingly

xuezhaojun: Got it.

PR #283, opened 2023-08-22
Remove dco check in github action.

xuezhaojun:
We already have a dco check in prow, and github action doc check is not always align with prow's result.

It also blocks this PR from redaht TAP. https://github.com/stolostron/managedcluster-import-controller/pull/231


zhujian7:
@qiujian16 WDYT?

xuezhaojun:
The latest blocked PR: https://github.com/stolostron/managedcluster-import-controller/pull/285

PR #284, opened 2023-08-22
preserve created-via hypershift annotation

rokej:
This is part of https://issues.redhat.com/browse/ACM-6547 to preserve `open-cluster-management/created-via: hypershift` annotation which is set by the hypershift hosted cluster auto-import controller.


zhujian7:
Hi @rokej, I saw there is a Train6 label on the story, so I will hold the PR until Train6 starts.
/hold

PR #286, opened 2023-08-27
Add KlusterletConfig support in bootstrap pkg.

xuezhaojun:
Fix: https://issues.redhat.com/browse/ACM-2252

### DIFF
--- a/pkg/helpers/helpers.go
+++ b/pkg/helpers/helpers.go
@@ -593,7 +593,7 @@ func GetComponentNamespace() (string, error) {
 	return string(nsBytes), nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/helpers/helpers.go
```diff
@@ -631,10 +627,6 @@ func GetTolerations(clusterAnnotations map[string]string) ([]corev1.Toleration,
 		return nil, fmt.Errorf("invalid tolerations annotation %v", err)
 	}
 
-	if err := validateTolerations(tolerations); err != nil {
```
xuezhaojun, line 634: Move validate out because klusterletconfig part needs validation too.

PR #287, opened 2023-08-28
Add Klusterletconfig in agent-registraiton and controller.

xuezhaojun:
Fix: https://issues.redhat.com/browse/ACM-2252

This part of code should shared with another Epic: proxy configuration. https://issues.redhat.com/browse/ACM-6325 These 2 Epics share the same tool code. cc @elgnay

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -25,12 +25,19 @@ import (
 	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
 	"github.com/stolostron/managedcluster-import-controller/pkg/controller"
 	"github.com/stolostron/managedcluster-import-controller/pkg/controller/agentregistration"
+	"github.com/stolostron/managedcluster-import-controller/pkg/controller/importconfig"
 	"github.com/stolostron/managedcluster-import-controller/pkg/features"
 	"github.com/stolostron/managedcluster-import-controller/pkg/helpers"
 	"github.com/stolostron/managedcluster-import-controller/pkg/helpers/imageregistry"
 	"github.com/stolostron/managedcluster-import-controller/pkg/source"
 
+	klusterletconfigclient "github.com/stolostron/cluster-lifecycle-api/client/klusterletconfig/clientset/versioned"
+	klusterletconfiginformerv1alpha1 "github.com/stolostron/cluster-lifecycle-api/client/klusterletconfig/informers/externalversions/klusterletconfig/v1alpha1"
+	listerklusterletconfigv1alpha1 "github.com/stolostron/cluster-lifecycle-api/client/klusterletconfig/listers/klusterletconfig/v1alpha1"
+	klusterletconfigv1alpha1 "github.com/stolostron/cluster-lifecycle-api/klusterletconfig/v1alpha1"
 	addonv1alpha1 "open-cluster-management.io/api/addon/v1alpha1"
+	clusterclient "open-cluster-management.io/api/client/cluster/clientset/versioned"
+	informersclusterv1 "open-cluster-management.io/api/client/cluster/informers/externalversions/cluster/v1"
 	operatorclient "open-cluster-management.io/api/client/operator/clientset/versioned"
 	workclient "open-cluster-management.io/api/client/work/clientset/versioned"
 	informersworkv1 "open-cluster-management.io/api/client/work/informers/externalversions/work/v1"
@@ -74,6 +81,7 @@ func init() {
 	utilruntime.Must(clusterv1.AddToScheme(scheme))
 	utilruntime.Must(asv1beta1.AddToScheme(scheme))
 	utilruntime.Must(addonv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(klusterletconfigv1alpha1.AddToScheme(scheme))
 }
 
 func main() {
@@ -126,6 +134,18 @@ func main() {
 	workClient, err := workclient.NewForConfig(cfg)
 	if err != nil {
 		setupLog.Error(err, "failed to create work client")
+		os.Exit(1)
+	}
+
+	klusterletconfigClient, err := klusterletconfigclient.NewForConfig(cfg)
+	if err != nil {
+		setupLog.Error(err, "failed to create klusterletconfig client")
+		os.Exit(1)
+	}
+
+	managedclusterClient, err := clusterclient.NewForConfig(cfg)
+	if err != nil {
+		setupLog.Error(err, "failed to create managedcluster client")
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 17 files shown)


cmd/manager/main.go
```diff
@@ -193,6 +202,9 @@ func main() {
 		},
 	)
 
+	klusterletconfigInformer := klusterletconfiginformer.NewSharedInformerFactory(
```
xuezhaojun, line 205: Inspired by this PR: https://github.com/open-cluster-management-io/ocm/pull/232 We choose to use `InformerFacotory` here.

cmd/manager/main.go
```diff
@@ -228,6 +241,8 @@ func main() {
 			KlusterletWorkLister:     listersworkv1.NewManifestWorkLister(klusterletWorksInformer.GetIndexer()),
 			HostedWorkInformer:       hostedWorksInformer,
 			HostedWorkLister:         listersworkv1.NewManifestWorkLister(hostedWorksInformer.GetIndexer()),
+			KlusterletConfigInformer: klusterletconfigInformer,
```
zhujian7, line 244: I did not see these fields used by any controllers or pkg in this PR.

zhujian7: Will there be another PR to modify the controllers or pkg to use these?

xuezhaojun: Yes, exactly, I try to split these into smaller PRs for better review.

zhujian7:
cc @zhiweiyin318

cmd/manager/main.go
xuezhaojun: After a thought, it's better to make it a complete submit. Change status to WIP.

pkg/controller/agentregistration/server.go
```diff
@@ -45,17 +50,35 @@ func RunAgentRegistrationServer(ctx context.Context, port int, clientHolder *hel
...
+		var kc *klusterletconfigv1alpha1.KlusterletConfig
+		klusterletconfigName := r.URL.Query().Get("klusterletconfig")
+		if klusterletconfigName != "" {
+			kc, err = klusterletconfigLister.Get(klusterletconfigName)
+			if err != nil && !apierrors.IsNotFound(err) {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+
+			klusterletClusterAnnotations["agent.open-cluster-management.io/klusterlet-config"] = klusterletconfigName
```
xuezhaojun, line 76: All places use `agent.open-cluster-management.io/klusterlet-config` will be replaced with a constant value after this PR merged: https://github.com/stolostron/cluster-lifecycle-api/pull/18

xuezhaojun: Except for calling `WithKlusterletConfig(kc)`, agent-registration also needs to add this annotation to `klusterletClusterAnnotations` to make sure agent-created managedcluster CR has this annotation. 

So that in the future reconcile loop in manifestwork-controller, the manifests keeps the same.

pkg/controller/importconfig/klusterletconfighandler.go
```diff
@@ -0,0 +1,68 @@
...
+			Name: mc.GetName(),
+		}})
+	}
+}
+
+const (
+	ManagedClusterKlusterletConfigAnnotationIndexKey = "annotation-klusterletconfig"
+)
+
+func IndexManagedClusterByKlusterletconfigAnnotation(obj interface{}) ([]string, error) {
```
xuezhaojun, line 58: @qiujian16  Please take a look if this index logic is what we talked about before, thanks!

xuezhaojun:
@qiujian16 @elgnay 
I'm think creating a default klusterletconfig for every newly created managedcluster should be handled by the ocm-webhook right? Or we should have in the import-controller repo?

Right now, if the klusterletconfig not specified in the annotations or not found, we just simply skip it. NO Default klusterletconfig yet.

elgnay:
@xuezhaojun I talked with @qiujian16 yesterday. We both agree that we're not going to create a default klusterletconfig cr, because once user changes the default klusterletconfig, it's hard for us to handle the configuration merge in the upgrading case.

xuezhaojun:
> @xuezhaojun I talked with @qiujian16 yesterday. We both agree that we're not going to create a default klusterletconfig cr, because once user changes the default klusterletconfig, it's hard for us to handle the configuration merge in the upgrading case.

Got it.

xuezhaojun:
/assign @qiujian16
All checks passed, PTAL, thanks!

cmd/manager/main.go
```diff
@@ -228,6 +264,9 @@ func main() {
...
+			KlusterletConfigLister:   listerklusterletconfigv1alpha1.NewKlusterletConfigLister(klusterletWorksInformer.GetIndexer()),
```
qiujian16, line 268: curious, why we need to set an indexer, why can't we get lister from informer?

pkg/controller/importconfig/manager.go
```diff
@@ -91,6 +93,18 @@ func Add(mgr manager.Manager, clientHolder *helpers.ClientHolder, informerHolder
 				UpdateFunc:  func(e event.UpdateEvent) bool { return true },
 			}),
 		).
+		Watches(
+			&klusterletconfigv1alpha1.KlusterletConfig{},
+			&enqueueManagedClusterInKlusterletConfigAnnotation{
+				managedclusterIndexer: informerHolder.ManagedClusterInformer.Cluster().V1().ManagedClusters().Informer().GetIndexer(),
+			},
+			builder.WithPredicates(predicate.Funcs{
```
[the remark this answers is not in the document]
qiujian16: yes I think so

cmd/manager/main.go
```diff
@@ -228,6 +264,9 @@ func main() {
 			KlusterletWorkLister:     listersworkv1.NewManifestWorkLister(klusterletWorksInformer.GetIndexer()),
 			HostedWorkInformer:       hostedWorksInformer,
 			HostedWorkLister:         listersworkv1.NewManifestWorkLister(hostedWorksInformer.GetIndexer()),
+			KlusterletConfigInformer: klusterletconfigInformer,
```
qiujian16, line 267: it is better to check if infomer has synced before starting the reconcile.

cmd/manager/main.go
xuezhaojun: Good point, it's just a style copy from `KlusterletWorkLister` and  `HostedWorkLister`, now look back it seems like a redudent.

Removed.

cmd/manager/main.go
xuezhaojun: Agree, added `WaitForCacheSync` for new informers.

xuezhaojun:
After a think, decide to use the same style to create informers.
The refactor to use informerFactory would go with a follow-up PR.

qiujian16:
is there any reason on not using informerfactory? It should be much easier to use

xuezhaojun:
> is there any reason on not using informerfactory? It should be much easier to use

Just want to keep a same style with previous code, I will submit a PR to refactor all informers to informerfactory very soon.

PR #289, opened 2023-09-01
support klusterletconfig with proxyconfig

elgnay:
The related issue: https://issues.redhat.com/browse/ACM-6325


xuezhaojun:
We may need to upgrade crd yaml used in e2e in this PR as well: cd9b187cca82

elgnay:
> @elgnay We may need to upgrade crd yaml used in e2e in this PR as well: https://github.com/stolostron/managedcluster-import-controller/blob/main/deploy/base/config.open-cluster-management.io_klusterletconfig.crd.yaml

Upgraded.

PR #290, opened 2023-09-04
Refactor to use informerFactory.

xuezhaojun:
A follow up of this: https://github.com/stolostron/managedcluster-import-controller/pull/287#issuecomment-1700581833

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -20,7 +20,9 @@ import (
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
+	"k8s.io/client-go/informers"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
 	"github.com/stolostron/managedcluster-import-controller/pkg/controller"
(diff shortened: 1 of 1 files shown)


xuezhaojun:
There is a trap I failed into that I want to share here:

https://github.com/kubernetes/client-go/blob/4783d283596145dd621e01636a829ec43b5101f9/informers/factory.go#L240

factory.Start() must be called before `ForResource` `Lister()` `Informer()` function. Otherwise we need to call Start again.

I once use the following code to start the agent-registration:
```go
        klusterletconfigInformerF.Start(ctx.Done())
	klusterletconfigInformerF.WaitForCacheSync(ctx.Done())

	managedclusterInformerF.Start(ctx.Done())
	managedclusterInformerF.WaitForCacheSync(ctx.Done())

	// Start the agent-registratioin server
	if features.DefaultMutableFeatureGate.Enabled(features.AgentRegistration) {
		go func() {
			if err := agentregistration.RunAgentRegistrationServer(ctx, 9091, clientHolder,
				klusterletconfigInformerF.Config().V1alpha1().KlusterletConfigs().Lister()); err != nil {
				setupLog.Error(err, "failed to start agent registration server")
			}
		}()
	}
```

Here agent-regisgtration server should wait for informers start then handle requests. BUT, when we call `klusterletconfigInformerF.Config().V1alpha1().KlusterletConfigs().Lister()`, under the hood, it creates a new informer which status is not running. 

So the correct way is:
```go
       klusterletconfigLister := klusterletconfigInformerF.Config().V1alpha1().KlusterletConfigs().Lister()
        ...
        klusterletconfigInformerF.Start(ctx.Done())
	klusterletconfigInformerF.WaitForCacheSync(ctx.Done())

	managedclusterInformerF.Start(ctx.Done())
	managedclusterInformerF.WaitForCacheSync(ctx.Done())

	// Start the agent-registratioin server
	if features.DefaultMutableFeatureGate.Enabled(features.AgentRegistration) {
		go func() {
			if err := agentregistration.RunAgentRegistrationServer(ctx, 9091, clientHolder,
				klusterletconfigLister); err != nil {
				setupLog.Error(err, "failed to start agent registration server")
			}
		}()
	}
```

Create `klusterletconfigLister` before `Start(context)`.

PR #294, opened 2023-09-20
Running singleton hosted mode by default

qiujian16:


qiujian16:
@Kyl-Bempah any idea why RHTAP fails?

Kyl-Bempah:
It's probably out of date since https://github.com/stolostron/managedcluster-import-controller/pull/231 has so many changes that it needs to merge

pkg/bootstrap/render.go
```diff
@@ -137,7 +137,7 @@ func (b *KlusterletManifestsConfig) Generate(ctx context.Context, clientHolder *
 	// Files depends on the install mode
 	var files []string
 	switch b.InstallMode {
-	case operatorv1.InstallModeHosted:
+	case operatorv1.InstallModeHosted, operatorv1.InstallModeSingletonHosted:
```
zhujian7, line 140: the mode value comes from https://github.com/stolostron/managedcluster-import-controller/blob/f92eb22264afeec6dcd485a313f79dd8733476d4/pkg/helpers/helpers.go#L634 , it seems that it is not possible to be `operatorv1.InstallModeSingletonHosted`

qiujian16: hrm I thought we will set the mode in annotation to be hosted. We do not need user to change the annotation to singletonhosted.

qiujian16:
hrm is this due to flaky e2e test?

PR #302, opened 2024-01-12
import rosa cluster with autoimportsecret

skeeey:


skeeey:
/hold

I need test this in real env

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -170,3 +193,31 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
+			getter = helpers.NewRosaKubeConfigGetter()
+			r.rosaKubeConfigGetters[string(rosaClusterID)] = getter
+		}
+
+		return func(secret *corev1.Secret) (reconcile.Result, *helpers.ClientHolder, meta.RESTMapper, error) {
+			return helpers.GenerateImportClientFromRosaCluster(getter, secret)
+		}, nil
+	}
+
+	return nil, fmt.Errorf("kubeconfig or token/server pair or rosa cluster info is required")
```
qiujian16, line 222: should return an import faiiled condition.

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -170,3 +193,31 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
 
 	return result, iErr
 }
+
+func (r *ReconcileAutoImport) getGenerateClientHolderFuncFromAutoImportSecret(secret *corev1.Secret) (helpers.GenerateClientHolderFunc, error) {
+	if _, hasKubeConfig := secret.Data["kubeconfig"]; hasKubeConfig {
```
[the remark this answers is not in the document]
skeeey: yes, I think we can,  I will try to update this part

pkg/helpers/helpers.go
```diff
@@ -109,82 +110,124 @@ func GetMaxConcurrentReconciles() int {
...
+			return reconcile.Result{}, nil, nil, err
 		}
+		return buildImportClient(config)
 	}
 
+	return reconcile.Result{}, nil, nil, fmt.Errorf("kubeconfig is missing")
+}
+
+// GenerateImportClientFromKubeTokenSecret generate a client from a given secret that contains kube apiserver and token
+func GenerateImportClientFromKubeTokenSecret(secret *corev1.Secret) (reconcile.Result, *ClientHolder, meta.RESTMapper, error) {
```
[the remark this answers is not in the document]
skeeey: we support using kube token + api server api  to import a cluster, in this case, we skip the tls verify...

https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.9/html/clusters/cluster_mce_overview#importing-clusters-auto-import-secret

pkg/helpers/helpers.go
```diff
@@ -109,82 +110,124 @@ func GetMaxConcurrentReconciles() int {
 	return maxConcurrentReconciles
 }
 
-// GenerateClientFromSecret generate a client from a given secret
-func GenerateClientFromSecret(secret *corev1.Secret) (*ClientHolder, meta.RESTMapper, error) {
```
zhujian7, line 113: the old `GenerateClientFromSecret` uses kubeconfig OR server/token, but after changing, the `GenerateImportClientFromKubeConfigSecret` will only use kubeconfig?

so the callers like `pkg/controller/clusterdeployment/clusterdeployment_controller.go` will be impacted?

pkg/controller/autoimport/autoimport_controller.go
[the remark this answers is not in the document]
skeeey: updated, using secret type to distinguish different import config

pkg/controller/autoimport/autoimport_controller.go
skeeey: If failed, we have a condition to represent this error, https://github.com/stolostron/managedcluster-import-controller/pull/302/files#diff-03d9e6ab77a9fa69840a6e5dfd9f090260a3c4707ce3d5c6f7aae0502663ea4cR142

pkg/helpers/rosa.go
```diff
@@ -0,0 +1,330 @@
...
+)
+
+/* #nosec */
+const (
+	defaultAPIServerURL = "https://api.openshift.com"
+	defaultTokenURL     = "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"
+
+	clusterAdminGroup = "cluster-admins"
+
+	importHTPasswdIDProvider = "acm-import"
```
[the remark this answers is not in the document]
skeeey: htpasswd is used to create and update the flat-files used to store usernames and password for basic authentication of HTTP users. 

HT may means HTTP, I think

pkg/helpers/helpers.go
```diff
@@ -109,82 +110,124 @@ func GetMaxConcurrentReconciles() int {
...
+	if retryTimes, ok := secret.Data["ocm_retry_times"]; ok {
+		getter.SetRetryTimes(string(retryTimes))
+	}
+
+	requeue, config, err := getter.KubeConfig()
+	if err != nil {
+		return reconcile.Result{Requeue: requeue, RequeueAfter: rosaImportRetryPeriod},
+			nil,
+			nil,
+			fmt.Errorf("failed to get kubeconfig for rosa cluster %s, retry after 1 minute", clusterID)
```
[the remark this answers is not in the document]
skeeey: corrected

pkg/helpers/helpers.go
skeeey: yes, handled by https://github.com/stolostron/managedcluster-import-controller/pull/302/files#diff-5fc238b39e2a1fc17bd412fbac2d14eab4e0f205c91886b5671611bec4c665ccR51

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -133,6 +137,25 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
 	}
 
+	generateClientHolderFunc, err := r.getGenerateClientHolderFuncFromAutoImportSecret(managedClusterName, autoImportSecret)
+	if err != nil {
+		if err := helpers.UpdateManagedClusterStatus(
+			r.client,
+			managedClusterName,
+			helpers.NewManagedClusterImportSucceededCondition(
+				metav1.ConditionFalse,
+				constants.ConditionReasonManagedClusterImportFailed,
```
qiujian16, line 147: I thought we already have an imported condition?

skeeey: currently, we use different condition to represent the cluster import process
https://github.com/stolostron/managedcluster-import-controller/blob/main/pkg/constants/constants.go#L108

pkg/helpers/autoimport.go
```diff
@@ -163,9 +162,20 @@ func (i *ImportHelper) Import(backupRestore bool, clusterName string,
 			), false, currentRetry, nil
 	}
 
-	clientHolder, restMapper, err := i.generateClientHolderFunc(managedClusterKubeClientSecret)
+	// build import client with managed cluster kube client secret
+	result, clientHolder, restMapper, err := i.generateClientHolderFunc(managedClusterKubeClientSecret)
```
zhujian7, line 166: since we removed the default `generateClientHolderFunc` in `NewImportHelper`, could this be nil?

pkg/helpers/helpers.go
zhujian7: > This because for clusterdeployment, it only have kubeconfig, right?

not sure about this, we need to confirm. Or can we just keep the origin logic? read kubeconfig first, if it does not exist, read token.

skeeey: we read the kubeconfig from `clusterdeployment.spec.clusterMetadata.adminKubeconfigSecretRef`, it  should only have kubeconfig, https://github.com/openshift/hive/blob/master/apis/hive/v1/clusterprovision_types.go#L44`

pkg/helpers/autoimport.go
skeeey: it could be ...., if this is nil, it must be a programmer level error, to avoid this add the check at the beginning

zhujian7:
LGTM

PR #309, opened 2024-01-31
add rosa import doc

skeeey:


docs/rosa_cluster_import.md
```diff
@@ -0,0 +1,50 @@
...
+
+1. Create a ManagedCluster on the hub cluster
+
+```sh
+oc apply -f - <<EOF
+apiVersion: cluster.open-cluster-management.io/v1
+kind: ManagedCluster
+metadata:
+  name: <your_cluster_name>
+  local-cluster: "true"
```
qiujian16, line 22: typo?

docs/rosa_cluster_import.md
```diff
@@ -0,0 +1,50 @@
...
+EOF
+```
+
+There are three optional options that can be added to secret for development purposes
+
+- `api_url`, The OpenShift API URL, the default value is https://api.openshift.com, it can be set to https://api.integration.openshift.com or https://api.staging.openshift.com
+- `token_url`, OpenID token URL. the default value is https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token
+- `retry_times`, The number of retries to obtain the ROSA cluster kube token, the default value is 20. The interval between each retry is 30 seconds.
+
+**Note**: The import controller will create a temporary cluster admin user `acm-import` with htPasswdIDProvider for your cluster, the import controller will use this user to fetch your cluster kube token and use this token to deploy the Klusterlet in your cluster. After your cluster is imported, the import controller will delete this temporary user.
```
qiujian16, line 50: should also mention the id provider name is hard coded. And when delete, both user and idprovider will be deleted.

docs/rosa_cluster_import.md
skeeey: yes, it is, removed

docs/rosa_cluster_import.md
skeeey: updated

PR #310, opened 2024-02-04
support klusterlet-priority-class

elgnay:
The related issue: https://issues.redhat.com/browse/ACM-6265

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -136,3 +136,7 @@ const (
 	AutoImportSecretRosaConfigClusterIDKey  string            = "cluster_id"
 	AutoImportSecretRosaConfigRetryTimesKey string            = "retry_times"
 )
+
+const (
+	DefaultKlusterletPriorityClassName = "klusterlet-critical"
+)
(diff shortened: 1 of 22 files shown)


pkg/helpers/helpers.go
```diff
@@ -62,8 +64,9 @@ import (
 const maxConcurrentReconcilesEnvVarName = "MAX_CONCURRENT_RECONCILES"
 
 const (
-	nodeSelectorAnnotation = "open-cluster-management/nodeSelector"
-	tolerationsAnnotation  = "open-cluster-management/tolerations"
+	nodeSelectorAnnotation      = "open-cluster-management/nodeSelector"
+	tolerationsAnnotation       = "open-cluster-management/tolerations"
+	kubeVersionClusterClaimName = "kubeversion.open-cluster-management.io"
```
qiujian16, line 69: you can actually get kube verison directly from status of managed cluster, right?

elgnay: Yes, I've updated the code as suggested.

pkg/constants/constants.go
```diff
@@ -136,3 +136,7 @@ const (
 	AutoImportSecretRosaConfigClusterIDKey  string            = "cluster_id"
 	AutoImportSecretRosaConfigRetryTimesKey string            = "retry_times"
 )
+
+const (
+	DefaultKlusterletPriorityClassName = "multicluster-engine-agent"
```
qiujian16, line 141: should we just call it klusterlet?

elgnay: Renamed.

PR #318, opened 2024-03-13
elevate permission of work agent to create priorityclass

elgnay:
The related issue: https://issues.redhat.com/browse/ACM-10267

### DIFF
--- a/pkg/bootstrap/manifests/klusterlet/clusterrole_priority_class.yaml
+++ b/pkg/bootstrap/manifests/klusterlet/clusterrole_priority_class.yaml
@@ -1,0 +1,10 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: klusterlet-priority-class
+  labels:
+    open-cluster-management.io/aggregate-to-work: "true"
+rules:
+- apiGroups: ["scheduling.k8s.io"]
+  resources: ["priorityclasses"]
+  verbs: ["get", "create", "patch", "update", "delete"]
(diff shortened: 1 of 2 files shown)


pkg/bootstrap/manifests/klusterlet/clusterrole_priority_class.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
zhujian7, line 1: Add a todo/note that we will remove this after 2 releases?

elgnay: Updated.

pkg/bootstrap/manifests/klusterlet/clusterrole_priority_class.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterRole
+metadata:
+  name: klusterlet-priority-class
+  labels:
+    open-cluster-management.io/aggregate-to-work: "true"
+rules:
+- apiGroups: ["scheduling.k8s.io"]
+  resources: ["priorityclasses"]
+  resourceNames: ["klusterlet-critical"]
```
zhujian7, line 10: we may need to remove the resourceNames, see the k8s doc [here](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources):
> Note: You cannot restrict create or deletecollection requests by their resource name. For create, this limitation is because the name of the new object may not be known at authorization time.

elgnay: Updated

PR #322, opened 2024-03-19
support deploy on non-ocp cluster

zhiweiyin318:
issue: https://issues.redhat.com/browse/ACM-10655

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -20,10 +20,6 @@ import (
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
-	"k8s.io/client-go/informers"
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/client-go/tools/cache"
-
 	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
 	"github.com/stolostron/managedcluster-import-controller/pkg/controller"
 	"github.com/stolostron/managedcluster-import-controller/pkg/controller/agentregistration"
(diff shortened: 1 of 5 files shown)


pkg/helpers/checkocp.go
```diff
@@ -0,0 +1,12 @@
+package helpers
+
+// globalDeployOnOCP is set once at the beginning
```
zhujian7, line 3:
```suggestion
// globalDeployOnOCP is set once at the beginning
var globalDeployOnOCP *bool

func SetDeployOnOCP(deployOnOcp bool) {
        if globalDeployOnOCP == nil
	    globalDeployOnOCP = deployOnOcp
	}
}
```

pkg/bootstrap/boostrapkubeconfig.go
```diff
@@ -169,12 +169,21 @@ func GetKubeAPIServerAddress(ctx context.Context, client client.Client,
 		return klusterletConfig.Spec.HubKubeAPIServerURL, nil
 	}
 
+	if !helpers.DeployOnOCP() {
+		return "", fmt.Errorf("failed get Hub kube apiserver on non-OCP cluster")
```
zhujian7, line 173: so here we return an error if it is not ocp, what function will be affected on non-ocp cluster?

zhiweiyin318: the bootstrap secret cannot be generated if cannot get hubkubeAPIServerURL here.

cmd/manager/main.go
```diff
@@ -84,7 +83,10 @@ func init() {
 
 func main() {
 	var leaderElectionNamespace string = ""
+	var deployOnOCP bool = true
 	pflag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "", "required when the process is not running in cluster")
+	pflag.BoolVar(&deployOnOCP, "deploy-on-ocp", true, "used to deploy the controller on OCP or not")
```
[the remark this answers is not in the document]
zhiweiyin318: good suggestion. 
done.

pkg/bootstrap/boostrapkubeconfig.go
zhujian7: so there will be no bootstrap secret on non-ocp cluster?

zhiweiyin318: bootstrap secret  is also needed in non-ocp cluster. And we only get hubkubeAPIServerURL from klusterletConfig in non-ocp cluster .

pkg/bootstrap/boostrapkubeconfig.go
```diff
@@ -184,6 +193,11 @@ func GetBootstrapCAData(ctx context.Context, clientHolder *helpers.ClientHolder,
 		return klusterletConfig.Spec.HubKubeAPIServerCABundle, nil
 	}
 
+	// get caBundle from the kube-root-ca.crt configmap in the pod namespace for non-ocp case.
```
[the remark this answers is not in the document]
zhiweiyin318: checked in AKS with 1.28.3 k8s,  it has kube-root-ca.crt configmap.
otherwise, we also can configure ca via klustereletconfig as long as the hub apiserver URL.

PR #327, opened 2024-04-03
✨ Emit cluster importing events

zhujian7:
https://issues.redhat.com/browse/ACM-10730

DDR: https://docs.google.com/document/d/1clRX2Lkdx6-GnZGPwTmyMtjaarj9EghS_GFYWuXpcEA/edit

### DIFF
--- a/pkg/controller/autoimport/autoimport_controller.go
+++ b/pkg/controller/autoimport/autoimport_controller.go
@@ -36,7 +36,7 @@ type ReconcileAutoImport struct {
 	client                client.Client
 	kubeClient            kubernetes.Interface
 	informerHolder        *source.InformerHolder
-	globalRecorder        events.Recorder
+	recorder              events.Recorder
 	mcRecorder            kevents.EventRecorder
 	importHelper          *helpers.ImportHelper
 	rosaKubeConfigGetters map[string]*helpers.RosaKubeConfigGetter
@@ -46,16 +46,16 @@ func NewReconcileAutoImport(
 	client client.Client,
 	kubeClient kubernetes.Interface,
 	informerHolder *source.InformerHolder,
-	globalRecorder events.Recorder,
+	recorder events.Recorder,
 	mcRecorder kevents.EventRecorder,
 ) *ReconcileAutoImport {
 	return &ReconcileAutoImport{
 		client:                client,
 		kubeClient:            kubeClient,
 		informerHolder:        informerHolder,
-		globalRecorder:        globalRecorder,
+		recorder:              recorder,
 		mcRecorder:            mcRecorder,
-		importHelper:          helpers.NewImportHelper(informerHolder, globalRecorder, log),
+		importHelper:          helpers.NewImportHelper(informerHolder, recorder, log),
 		rosaKubeConfigGetters: make(map[string]*helpers.RosaKubeConfigGetter),
 	}
 }
@@ -105,6 +105,7 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	reqLogger.V(5).Info("Reconciling auto import secret")
+
 	lastRetry := 0
 	totalRetry := 1
 	if len(autoImportSecret.Annotations[constants.AnnotationAutoImportCurrentRetry]) != 0 {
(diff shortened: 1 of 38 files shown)


pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -102,7 +101,7 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	reqLogger.V(5).Info("Reconciling auto import secret")
-
+	mcEventRecorder := helpers.NewManagedClusterEventRecorder(r.kubeClient, controllerName, managedCluster)
```
zhujian7, line 104: TODO: need to shut down the event recorder after ensuring the event is reported. Otherwise, this may result in a memory leak.

zhujian7: using the Kubernetes client-go event library can resolve this issue.

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -44,15 +46,16 @@ func NewReconcileAutoImport(
 	client client.Client,
 	kubeClient kubernetes.Interface,
 	informerHolder *source.InformerHolder,
-	recorder events.Recorder,
+	globalRecorder events.Recorder,
+	mcRecorder kevents.EventRecorder,
```
zhiweiyin318, line 50: it is difficult to understand why we need 2 recorder here if there is no comments.
do we consider to merge the two recorders ?  
the events are in different namespaces, how about use one method here and diff them in the method ?

pkg/controller/managedcluster/managedcluster_controller.go
```diff
@@ -96,6 +110,26 @@ func (r *ReconcileManagedCluster) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, nil
 	}
 
+	// add a detaching condition to the managed cluster if the managed cluster is deleting
+	// if it is already in detaching or force deaching state, skip it
+	ic := meta.FindStatusCondition(managedCluster.Status.Conditions, constants.ConditionManagedClusterImportSucceeded)
```
elgnay, line 115: Do the importing and detaching share the same condition type with different reasons? why not use separated condition types?

zhujian7: yes, Importing and detaching represent a state of the life cycle. They have the same dimension and are mutually exclusive. It's easier to use the same type of condition

pkg/helpers/helpers.go
```diff
@@ -269,29 +271,79 @@ func RemoveManagedClusterFinalizer(ctx context.Context, runtimeClient client.Cli
...
+			"The %s is waiting for importing", mc.Name)
+	case constants.ConditionReasonManagedClusterImporting:
+		recorder.Eventf(mc, nil, corev1.EventTypeNormal,
+			constants.EventReasonManagedClusterImporting, constants.EventReasonManagedClusterImporting,
+			"The %s is being imported now: %s", mc.Name, cond.Message)
+	case constants.ConditionReasonManagedClusterImported:
+		recorder.Eventf(mc, nil, corev1.EventTypeNormal,
+			constants.EventReasonManagedClusterImported, constants.EventReasonManagedClusterImported,
+			"The %s is imported successfully", mc.Name)
+	case constants.ConditionReasonManagedClusterImportFailed:
```
elgnay, line 332: How many ImportFailed events will be emitted if the auto-import keeps failed.

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -41,13 +43,15 @@ func NewReconcileClusterDeployment(
...
 	recorder events.Recorder,
+	mcRecorder kevents.EventRecorder,
 ) *ReconcileClusterDeployment {
 
 	return &ReconcileClusterDeployment{
 		client:         client,
 		kubeClient:     kubeClient,
 		informerHolder: informerHolder,
 		recorder:       recorder,
+		mcRecorder:     mcRecorder,
```
qiujian16, line 54: why do we still need the original recorder?

pkg/helpers/helpers.go
zhujian7: it will be multiple events since the retry mechanism causes the condition message changes, like `Try to import managed cluster, retry times: 1/3, error: xxx`

pkg/controller/clusterdeployment/clusterdeployment_controller.go
zhujian7: yes, because we use the openshift library go to apply some resources, and the library go needs the openshift event recorder.

PR #330, opened 2024-04-09
Add "global" KlusterletConfig Support.

xuezhaojun:
Story: https://issues.redhat.com/browse/ACM-10074

image

We add global klusterletconfig related code in 2 places:
* agentregistration server
* import-config-controller

Currently we are only required to support 1 special "global" klustereletconfig, but in the future if we want to support multiple klusterletconfig, we need to refactor more.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -140,3 +140,7 @@ const (
 const (
 	DefaultKlusterletPriorityClassName = "klusterlet-critical"
 )
+
+const (
+	GlobalKlusterletConfigName = "global"
+)
--- a/pkg/helpers/klusterletconfig.go
+++ b/pkg/helpers/klusterletconfig.go
@@ -1,0 +1,32 @@
+package helpers
+
+import (
+	"fmt"
+
+	listerklusterletconfigv1alpha1 "github.com/stolostron/cluster-lifecycle-api/client/klusterletconfig/listers/klusterletconfig/v1alpha1"
+	klusterletconfighelper "github.com/stolostron/cluster-lifecycle-api/helpers/klusterletconfig"
+	klusterletconfigv1alpha1 "github.com/stolostron/cluster-lifecycle-api/klusterletconfig/v1alpha1"
+	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+func GetMergedKlusterletConfigWithGlobal(
+	klusterletconfigName string,
+	kcLister listerklusterletconfigv1alpha1.KlusterletConfigLister,
+) (*klusterletconfigv1alpha1.KlusterletConfig, error) {
+	var err error
+	var kc *klusterletconfigv1alpha1.KlusterletConfig
+	if klusterletconfigName != "" {
+		kc, err = kcLister.Get(klusterletconfigName)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get klusterletconfig %s: %v", klusterletconfigName, err)
+		}
+	}
+
+	globalKlusterletConfig, err := kcLister.Get(constants.GlobalKlusterletConfigName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get global klusterletconfig: %v", err)
+	}
+
+	return klusterletconfighelper.MergeKlusterletConfigs(globalKlusterletConfig, kc)
+}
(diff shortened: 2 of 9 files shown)


pkg/bootstrap/render.go
```diff
@@ -370,3 +377,51 @@ func imageOverride(source, mirror, imageName string) string {
 	trimSegment := strings.TrimPrefix(imageName, source)
 	return fmt.Sprintf("%s%s", mirror, trimSegment)
 }
+
+// Merge multiple klusterletconfigs into one klusterletconfigSpec.
+// For each field in klusterletconfigSpec, we take the first non-empty value from the klusterletconfigs list.
+func mergeKlusterletConfigSpec(klusterletconfigs []*klusterletconfigv1alpha1.KlusterletConfig) *klusterletconfigv1alpha1.KlusterletConfigSpec {
```
xuezhaojun, line 383: cc @elgnay

pkg/helpers/klusterletconfig.go
```diff
@@ -0,0 +1,32 @@
...
+		kc, err = kcLister.Get(klusterletconfigName)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get klusterletconfig %s: %v", klusterletconfigName, err)
+		}
+	}
+
+	globalKlusterletConfig, err := kcLister.Get(constants.GlobalKlusterletConfigName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get global klusterletconfig: %v", err)
+	}
```
xuezhaojun, line 29: If "not found" return kc directly without run merge. (@zhiwei)

pkg/helpers/klusterletconfig.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+
+	listerklusterletconfigv1alpha1 "github.com/stolostron/cluster-lifecycle-api/client/klusterletconfig/listers/klusterletconfig/v1alpha1"
+	klusterletconfighelper "github.com/stolostron/cluster-lifecycle-api/helpers/klusterletconfig"
+	klusterletconfigv1alpha1 "github.com/stolostron/cluster-lifecycle-api/klusterletconfig/v1alpha1"
+	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+func GetMergedKlusterletConfigWithGlobal(
```
xuezhaojun, line 13: Do we need to rename to GetKlusterletConfig.

pkg/helpers/klusterletconfig.go
```diff
@@ -0,0 +1,32 @@
...
+			return nil, fmt.Errorf("failed to get klusterletconfig %s: %v", klusterletconfigName, err)
+		}
+	}
+
+	globalKlusterletConfig, err := kcLister.Get(constants.GlobalKlusterletConfigName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to get global klusterletconfig: %v", err)
+	}
+
+	return klusterletconfighelper.MergeKlusterletConfigs(globalKlusterletConfig, kc)
```
qiujian16, line 31: do you support the nil input?

xuezhaojun: Yes, this PR https://github.com/stolostron/cluster-lifecycle-api/pull/25 will filter out the nil items.

pkg/helpers/klusterletconfig.go
xuezhaojun: Kindly note the name should retain the 'Global' part for the sake of readability.

pkg/constants/constants.go
```diff
@@ -140,3 +140,7 @@ const (
 const (
 	DefaultKlusterletPriorityClassName = "klusterlet-critical"
 )
+
+const (
+	GlobalKlusterletConfigName = "global"
```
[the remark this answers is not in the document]
xuezhaojun: I kind of want to keep global as part of import-controller because merging klusterletconfig seems essential, but the rule that we hold a special klusterletconfig named "global" could be modified.

pkg/helpers/klusterletconfig.go
xuezhaojun: The `klusterletconfighelper.MergeKlusterletConfigs` implement the process to filterout the nil items, it can save us some code from the client side.

xuezhaojun: I want in the future when we want to implement a real multiple kc merge process, we also don't need to process each kc is nil, just put all kcs as the input and the merge function will handle it.

PR #338, opened 2024-05-07
Customize klusterlet namespace from klusterletConfig

qiujian16:


pkg/bootstrap/render.go
```diff
@@ -329,6 +331,32 @@ func filesToObjects(files []string, config interface{}) ([]runtime.Object, error
...
+func getKlusterletName(config *klusterletconfigv1alpha1.KlusterletConfig, clusterName, namespace string, mode operatorv1.InstallMode) string {
+	if mode == operatorv1.InstallModeHosted || mode == operatorv1.InstallModeSingletonHosted {
+		return fmt.Sprintf("klusterlet-%s", clusterName)
+	}
+	klusterletNamespace := getKlusterletNamespace(namespace, config)
+	if klusterletNamespace == constants.DefaultKlusterletNamespace {
+		return "klusterlet"
+	}
+	postFix := strings.TrimPrefix(klusterletNamespace, "open-cluster-management-")
+	return fmt.Sprintf("klusterlet-%s", postFix)
```
zhujian7, line 357: so the klusterlet name is determined by klusterlet namespace?
if the klusterlet namespace changes, we will delete the old klusterlet and create a new one?

zhujian7: the klustelet manifestwork delete option is orphan, so if the klusterlet namespace changes, there will be 2 klusterlet CRs on the managed cluster, right? I do not think that is what we expected.

qiujian16:
before we merge this, do we also need to update mce installer with updated CRD at first @elgnay

elgnay:
> before we merge this, do we also need to update mce installer with updated CRD at first @elgnay

Yes, we cannot merge this PR until the KlusterletConfig API is updated in the backplane-operator repo.

PR #346, opened 2024-05-21
Use infra uid to generate bootstrap kubeconfig

zhujian7:
issue reference: https://issues.redhat.com/browse/ACM-10626


elgnay:
We may also need to extract the cluster name from the existing bootstrap hub kubeconfig here https://github.com/stolostron/managedcluster-import-controller/pull/346/files#diff-576c58639c3f314f0ece65ca6e525856a23b8d7087214a1653ca35071d07be48L42, and compare it with the UUID of the Infrastructure. Then determine whether we need to re-create the bootstrap hub kubeconfig.

zhujian7:
> We may also need to extract the cluster name from the existing bootstrap hub kubeconfig

@elgnay fixed, PTAL

pkg/bootstrap/boostrapkubeconfig.go
```diff
@@ -181,6 +187,26 @@ func GetKubeAPIServerAddress(ctx context.Context, client client.Client,
 	return "", err
 }
 
+func GetKubeconfigClusterName(ctx context.Context, client client.Client) (string, error) {
+	defaultCluster := "default-cluster"
+	if !helpers.DeployOnOCP {
```
qiujian16, line 192: We should add a comment here. We will miss the function if hub is non-ocp, but we are trying to support that.

pkg/controller/importconfig/cluster_info.go
```diff
@@ -74,6 +75,16 @@ func getBootstrapKubeConfigDataFromImportSecret(ctx context.Context, clientHolde
 		return nil, nil, nil
 	}
 
+	// check if the current context cluster name of the bootstrap kubeconfig is changed
+	validCtxClusterName, err := validateContextClusterName(ctx, clientHolder.RuntimeClient, ctxClusterName)
```
qiujian16, line 79: note: why we validate here? all generation and validation should be in bootstrap package.

qiujian16 (commented):
just some comment should be add. And we need consider some refactor.

pkg/controller/importconfig/cluster_info.go
zhujian7: this is to fix YangLe's comment https://github.com/stolostron/managedcluster-import-controller/pull/346#issuecomment-2126303203

pkg/bootstrap/boostrapkubeconfig.go
zhujian7: PTAL

pkg/controller/importconfig/cluster_info.go
qiujian16: yeah I see. I mean the whole validation func should be put in the bootstrap package.

zhujian7: I will submit a follow-up PR to refactor this, and keep the current PR changes as small as possible to facilitate cherrypick to the old releases.

PR #348, opened 2024-05-28
Move bootstap kubeconfig validation func to bootstrap pkg

zhujian7:
Fix comment: https://github.com/stolostron/managedcluster-import-controller/pull/346#discussion_r1616528184


xuezhaojun:
/lgtm

/hold 
if need double check.

PR #373, opened 2024-07-15
add token creation time in the import secret

zhiweiyin318:


zhujian7:
lgtm

pkg/controller/importconfig/cluster_info.go
```diff
@@ -123,13 +125,25 @@ func validateToken(token string, expiration []byte) bool {
 		return true
 	}
 
+	if len(creation) == 0 {
+		klog.Infof("token creation is empty, expiration: %v", string(expiration))
+		return false
```
elgnay, line 130: Once an ACM instance upgrades to this code, all existing bootstrap kubeconfig secret become invalid, which will result in restarting of the klusterlet-agent. Is there a way to avoid this?

zhiweiyin318: changed to use default refreshThreshold if creation is empty.

PR #374, opened 2024-07-15
Return short-term token in agent-regstration endpoint.

xuezhaojun:
Fix: https://issues.redhat.com/browse/ACM-12277

In this PR, we change agent-registration to return a short-term bootstrapkubeconfig secret if the `duration` param in the URL is set, for example:

```sh
curl --cacert ca.crt -H "Authorization: Bearer $token" [https://$agent_registration_host/agent-registration/manifests/cluster-e2e-test-agent?klusterletconfig\=default\&expiration\=4h](https://$agent_registration_host/agent-registration/manifests/cluster-e2e-test-agent?klusterletconfig=default&duration=4h)
```

### DIFF
--- a/pkg/controller/agentregistration/server.go
+++ b/pkg/controller/agentregistration/server.go
@@ -57,13 +57,14 @@ func RunAgentRegistrationServer(ctx context.Context, port int, clientHolder *hel
 		}
 	})))
 
-	// example URl: https://<route address>/agent-registration/manifests/cluster1?klusterletconfig=default
+	// example URl: https://<route address>/agent-registration/manifests/cluster1?klusterletconfig=default&duration=4h
 	mux.Handle("/agent-registration/manifests/", authMiddleware(clientHolder, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		var err error
 		urlparams := strings.Split(r.URL.Path, "/")
 		clusterID := urlparams[len(urlparams)-1]
 
 		klusterletconfigName := r.URL.Query().Get("klusterletconfig")
+		durationStr := r.URL.Query().Get("duration")
 
 		// Get the merged KlusterletConfig, it merges the user assigned KlusterletConfig with the global KlusterletConfig.
 		mergedKlusterletConfig, err := helpers.GetMergedKlusterletConfigWithGlobal(klusterletconfigName, klusterletconfigLister)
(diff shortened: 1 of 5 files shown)


pkg/controller/agentregistration/server.go
```diff
@@ -74,9 +75,30 @@ func RunAgentRegistrationServer(ctx context.Context, port int, clientHolder *hel
...
+				return
+			}
+		} else {
+			expiration, err := time.ParseDuration(expirationStr)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+
+			token, _, err = bootstrap.RequestSAToken(ctx, clientHolder.KubeClient, AgentRegistrationDefaultBootstrapSAName, ns, int64(expiration.Seconds()))
```
zhiweiyin318, line 94: double check: will request a new token every time?

zhiweiyin318 (commented):
/approve
need @zhujian7  take another look

pkg/controller/agentregistration/server.go
xuezhaojun: Not every time, only if the `expiration` in the url parameter is set.
Regularly, if the `expiration` is not set, it will get token from SA secret.

zhujian7: I am not sure about the case that the real expiration time returned by `bootstrap.RequestSAToken` is different from the expiration time the user requested(the case @zhiweiyin318 is working on), should we return the real expiration time to the user?

pkg/controller/agentregistration/server.go
```diff
@@ -74,9 +75,30 @@ func RunAgentRegistrationServer(ctx context.Context, port int, clientHolder *hel
...
 		// In the agent-registration case, the bootstrap sa is not created in the managed cluster namespace, because managed cluster is not created yet.
 		// Instead, it's in the pod namespace with the name "agent-registration-bootstrap".
-		bootstrapkubeconfig, _, err := bootstrap.CreateBootstrapKubeConfig(ctx, clientHolder, AgentRegistrationDefaultBootstrapSAName,
-			os.Getenv(constants.PodNamespaceEnvVarName),
-			7*24*3600, mergedKlusterletConfig)
+		ns := os.Getenv(constants.PodNamespaceEnvVarName)
+
+		var token []byte
+		if expirationStr == "" {
+			token, err = bootstrap.GetTokenFromSASecret(ctx, clientHolder.KubeClient, AgentRegistrationDefaultBootstrapSAName, ns)
```
zhujian7, line 82: I think we have some behavior changes here, before the PR, when the expirationStr is empty, if getting the token from the sa failed, it will use the tokenRequest to create a new token, but now, it will just return fail. Is that on purpose?

xuezhaojun: Yes, agent-registration is using a sa that always will always has the sa secret exists.

pkg/controller/agentregistration/server.go
xuezhaojun: Good point, I didn't look the PR https://github.com/stolostron/managedcluster-import-controller/pull/373 before, let me take a look now.

pkg/controller/agentregistration/server.go
xuezhaojun: This is because the agent-registration returns the manifests that may or may not applied on a cluster.

So I want the different URl matches a determined result:
* If the `expiration` is set, then token with that expiration.
* if the `expiration` is not set, then token has no expiration.

pkg/controller/agentregistration/server.go
xuezhaojun: The PR inspired me, the parameter name should be duration instead of expiration.

xuezhaojun: Just like kubectl, it use duration in the `create token` command:

```
kubectl create token myapp --duration 10m
```

xuezhaojun: Updated to `duration` in parameter.

xuezhaojun: Echo the comment from @zhujian7 , the request expiration may not equals to the actual expiration returns by issuer:
```go
	// ExpirationSeconds is the requested duration of validity of the request. The
	// token issuer may return a token with a different validity duration so a
	// client needs to check the 'expiration' field in a response.
	// +optional
	ExpirationSeconds *int64 `json:"expirationSeconds" protobuf:"varint,4,opt,name=expirationSeconds"`
```

pkg/controller/agentregistration/server.go
xuezhaojun: What I want to avoid is: user doesn't set the `duration`, he get the result successfully, but now he doesn't know if the bootstraokubeconfig return by acm is long-term or 7 days.

He has to decode the token to know that.

But yes,  this is a behivior change comparing with the previous version.

xuezhaojun:
@zhujian7 @zhiweiyin318 
PR rebased, please take another look, thanks!

PR #379, opened 2024-07-25
Upgrade go version and related dependences.

xuezhaojun:
In the process of upgrade I met the same issue:

```
/lib64/libc.so.6: version `GLIBC_2.34' not found
```

To solve this I updated `registry.access.redhat.com/ubi8/ubi-minimal:latest` to `registry.access.redhat.com/ubi9/ubi-minimal:latest` as this thread suggested: https://redhat-internal.slack.com/archives/C04L50S5XM4/p1721765040271779"

Also need to upgrade the base image used in prow config: https://github.com/openshift/release/pull/54852


Makefile
```diff
@@ -13,7 +13,7 @@ export ARCH_TYPE   = $(if $(patsubst x86_64,,$(ARCH)),$(ARCH),amd64)
 export BUILD_DATE  = $(shell date +%m/%d@%H:%M:%S)
 export VCS_REF     = $(if $(shell git status --porcelain),$(GIT_COMMIT)-$(BUILD_DATE),$(GIT_COMMIT))
 
-export CGO_ENABLED  = 1
```
xuezhaojun, line 16: In the process of upgrade I met the same issue of: https://redhat-internal.slack.com/archives/C02A1PX4B9S/p1686886726859979

```
/lib64/libc.so.6: version `GLIBC_2.34' not found
```

To solve this I updated the CGO_ENABLED by default to 0.

xuezhaojun: But do you remember why in the first place we change it to `1` by this PR https://github.com/stolostron/managedcluster-import-controller/pull/235 ? @elgnay

PR #382, opened 2024-08-06
Update Klusterlet CRD.

xuezhaojun:


pkg/bootstrap/manifests/klusterlet/crds/klusterlets.crd.v1beta1.yaml
```diff
@@ -131,11 +131,15 @@ spec:
                           type: integer
                           format: int32
                           minimum: 180
-                        secretNames:
-                          description: SecretNames is a list of secret names. The secrets are in the same namespace where the agent controller runs.
+                        kubeConfigSecrets:
```
zhujian7, line 134: In ACM, we have not used the registrationConfiguration, so this will not break the upgrading, right?

PR #383, opened 2024-08-07
keep the user role binding

skeeey:
to avoid https://issues.redhat.com/browse/OCM-9916, only delete the import user and keep the role binding after imported a rosa cluster


skeeey:
Close this as the root cause is found

PR #385, opened 2024-08-14
Add the support of "MultipleHubs" in KlusterletConfig.

xuezhaojun:
Fixes: https://issues.redhat.com/browse/ACM-11114

### DIFF
--- a/pkg/bootstrap/render.go
+++ b/pkg/bootstrap/render.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/stolostron/cluster-lifecycle-api/helpers/localcluster"
 	klusterletconfigv1alpha1 "github.com/stolostron/cluster-lifecycle-api/klusterletconfig/v1alpha1"
 	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
 	"github.com/stolostron/managedcluster-import-controller/pkg/helpers"
@@ -20,6 +21,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	apifeature "open-cluster-management.io/api/feature"
 	operatorv1 "open-cluster-management.io/api/operator/v1"
 )
 
@@ -67,12 +71,16 @@ type RenderConfig struct {
 	ImagePullSecretConfig
 }
 
+type BootstrapKubeConfigSecret struct {
+	Name       string
+	KubeConfig string
+}
+
 // KlusterletRenderConfig defines variables used in the klusterletFiles.
 type KlusterletRenderConfig struct {
 	KlusterletName            string
 	KlusterletNamespace       string
 	ManagedClusterNamespace   string
-	BootstrapKubeconfig       string
 	RegistrationOperatorImage string
 	RegistrationImageName     string
 	WorkImageName             string
@@ -86,6 +94,10 @@ type KlusterletRenderConfig struct {
 
 	RegistrationConfiguration *operatorv1.RegistrationConfiguration
 	WorkConfiguration         *operatorv1.WorkAgentConfiguration
+
+	MultipleHubsEnabled              bool
+	DefaultBootstrapKubeConfigSecret BootstrapKubeConfigSecret
+	BootstrapKubeConfigSecrets       []BootstrapKubeConfigSecret
 }
 
 type ImagePullSecretConfig struct {
@@ -107,8 +119,8 @@ type KlusterletManifestsConfig struct {
 	// PriorityClassName is the name of the PriorityClass used by the klusterlet and operator
 	PriorityClassName string
 
-	// In the managed cluster annotations, it contains nodeSelectors and tolerations for the klusterlet deployment.
-	ManagedClusterAnnotations map[string]string
+	// Used to determine whether mc is a localcluster.
+	ManagedCluster *clusterv1.ManagedCluster
 
 	klusterletconfig *klusterletconfigv1alpha1.KlusterletConfig
 
@@ -132,11 +144,9 @@ func (c *KlusterletManifestsConfig) WithKlusterletClusterAnnotations(a map[strin
 	return c
 }
 
-// WithManagedClusterAnnotations sets the managed cluster annotations.
-// The managed cluster annotations contains information like: image pull secret, nodeSelector, tolerations, etc.
-// We need to extract these information from the managed cluster annotations to render the klusterlet manifests.
-func (c *KlusterletManifestsConfig) WithManagedClusterAnnotations(a map[string]string) *KlusterletManifestsConfig {
-	c.ManagedClusterAnnotations = a
+// WithManagedClusterLabels sets the managed cluster.
+func (c *KlusterletManifestsConfig) WithManagedCluster(mc *clusterv1.ManagedCluster) *KlusterletManifestsConfig {
+	c.ManagedCluster = mc
 	return c
 }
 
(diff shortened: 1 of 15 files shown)


xuezhaojun:
cc @clyang82

pkg/bootstrap/render.go
```diff
@@ -319,6 +337,39 @@ func (b *KlusterletManifestsConfig) Generate(ctx context.Context, clientHolder *
...
+			multipleHubsEnabled = true
+			for _, s := range b.klusterletconfig.Spec.BootstrapKubeConfigs.LocalSecrets.KubeConfigSecrets {
+				ns := os.Getenv(constants.PodNamespaceEnvVarName)
+				secret, err := clientHolder.KubeClient.CoreV1().Secrets(ns).Get(ctx, s.Name, metav1.GetOptions{})
+				if err != nil {
+					return nil, err
+				}
+				bootstrapKubeConfigSecrets = append(bootstrapKubeConfigSecrets, BootstrapKubeConfigSecret{
+					Name:       s.Name,
+					KubeConfig: base64.StdEncoding.EncodeToString(secret.Data["kubeconfig"]),
```
zhujian7, line 364: should we check whether the "kubeconfig" key exists in the secret data?

pkg/bootstrap/render.go
```diff
@@ -319,6 +337,39 @@ func (b *KlusterletManifestsConfig) Generate(ctx context.Context, clientHolder *
...
+			registrationConfiguration.FeatureGates = append(registrationConfiguration.FeatureGates,
+				operatorv1.FeatureGate{
+					Feature: string(apifeature.MultipleHubs),
+					Mode:    operatorv1.FeatureGateModeTypeEnable,
+				})
+			registrationConfiguration.BootstrapKubeConfigs = b.klusterletconfig.Spec.BootstrapKubeConfigs
+
+			multipleHubsEnabled = true
+			for _, s := range b.klusterletconfig.Spec.BootstrapKubeConfigs.LocalSecrets.KubeConfigSecrets {
+				ns := os.Getenv(constants.PodNamespaceEnvVarName)
```
zhujian7, line 357: The secret can only be in the pod namespace? in MCE it is `multicluster-engine`?

pkg/controller/importconfig/manager.go
```diff
@@ -123,6 +123,19 @@ func Add(ctx context.Context,
 					UpdateFunc:  func(e event.UpdateEvent) bool { return true },
 				})),
 		).
+		Watches(
+			&corev1.Secret{},
```
zhujian7, line 127: will this watch all secrets in the hub cluster?

pkg/controller/importconfig/klusterletconfighandler.go
```diff
@@ -69,3 +71,81 @@ func IndexManagedClusterByKlusterletconfigAnnotation(obj interface{}) ([]string,
...
+	evt event.DeleteEvent, q workqueue.RateLimitingInterface) {
+	e.enqueue(evt.Object.GetName(), q)
+}
+
+func (e *enqueueManagedClusterByBootstrapKubeConfigSecrets) Generic(ctx context.Context,
+	evt event.GenericEvent, q workqueue.RateLimitingInterface) {
+	e.enqueue(evt.Object.GetName(), q)
+}
+
+func (e *enqueueManagedClusterByBootstrapKubeConfigSecrets) enqueue(secretName string, q workqueue.RateLimitingInterface) {
```
zhujian7, line 111: can we have some rules like the secret must be in the managed cluster namespace, so here we can enqueue the secret.namespace directly?

pkg/bootstrap/render.go
xuezhaojun: Yes, within the same namespace of the pod.

pkg/bootstrap/render.go
xuezhaojun: Sure, the check added.

pkg/controller/importconfig/manager.go
xuezhaojun: Good point, change to watch only secrets under pod namespace.

pkg/controller/importconfig/klusterletconfighandler.go
xuezhaojun: Not quit get the idea of enqueue the secret.namespace, may need to discuss offline.

pkg/bootstrap/render.go
```diff
@@ -319,6 +337,45 @@ func (b *KlusterletManifestsConfig) Generate(ctx context.Context, clientHolder *
...
+				secret, err := clientHolder.KubeClient.CoreV1().Secrets(ns).Get(ctx, s.Name, metav1.GetOptions{})
+				if err != nil {
+					return nil, err
```
qiujian16, line 360: do we really want to return error here? what it type is not local secrets and the secrets are specified here?

xuezhaojun:
whether “watch all secrets in mce ns” will cause a performance issue should be considered.

pkg/bootstrap/render.go
```diff
@@ -319,6 +337,45 @@ func (b *KlusterletManifestsConfig) Generate(ctx context.Context, clientHolder *
 		renderConfig.ImagePullSecretConfig = imagePullSecretConfig
 	}
 
+	// MultipleHubs
+	// Doesn't affect on the local-cluster.
+	// Using MultipleHubs can controls the bootstrap kubeconfig secret/secrets easier.
+	multipleHubsEnabled := false
+	bootstrapKubeConfigSecrets := []BootstrapKubeConfigSecret{}
+	if local, ok := b.ManagedClusterLabels[constants.SelfManagedLabel]; !ok || local != "true" {
```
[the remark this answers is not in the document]
xuezhaojun: Update.

pkg/bootstrap/render.go
```diff
@@ -319,6 +337,45 @@ func (b *KlusterletManifestsConfig) Generate(ctx context.Context, clientHolder *
 		renderConfig.ImagePullSecretConfig = imagePullSecretConfig
 	}
 
+	// MultipleHubs
+	// Doesn't affect on the local-cluster.
+	// Using MultipleHubs can controls the bootstrap kubeconfig secret/secrets easier.
+	multipleHubsEnabled := false
```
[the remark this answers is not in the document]
xuezhaojun: I don't quit understand the idea, the multiplehubs need to render the `Klusterlet`, it's not likely work like `GenerateKlusterletCRDsV1` which only need to render independent files.

pkg/bootstrap/render.go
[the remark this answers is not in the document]
xuezhaojun: Updated.

pkg/bootstrap/render.go
xuezhaojun: So if type is not localsecrets the path would go down here.

pkg/controller/importconfig/manager.go
```diff
@@ -123,6 +128,27 @@ func Add(ctx context.Context,
 					UpdateFunc:  func(e event.UpdateEvent) bool { return true },
 				})),
 		).
+		Watches(
```
zhujian7, line 131: Please also check if the import-controller serviceaccount in the backplane-operator has the permission.

qiujian16:
need rebase, otherwise looks good.

pkg/controller/importconfig/manager.go
xuezhaojun: Good point, checked: https://github.com/stolostron/backplane-operator/blob/0d2edc3cf537a2eb4c0c9af71e1a98d104110ca0/pkg/templates/charts/toggle/server-foundation/templates/managedcluster-import-role.yaml#L73C1-L86C10

xuezhaojun:
> need rebase, otherwise looks good.

Thanks! Code rebased.

xuezhaojun:
@qiujian16 @zhujian7  All checks are passed, looking for another round of reivew, thanks!

qiujian16:
/approve
@zhujian7 would you do the final review?

zhujian7:
/lgtm
/approve
@xuezhaojun please unhold if QE can contain this in this train.

PR #391, opened 2024-08-26
Update dockerfile to assume yes in microdnf update process.

xuezhaojun:


zhiweiyin318:
I think we need update rhtap dockerfile too.

PR #393, opened 2024-09-03
fix create-via annotation not update

zhiweiyin318:
issue: https://issues.redhat.com/browse/ACM-12756


zhujian7:
can you add an integration/e2e test?

PR #395, opened 2024-09-11
fix only update cluster url after apply resources

zhiweiyin318:
issue: https://issues.redhat.com/browse/ACM-13722


pkg/controller/autoimport/apiserverurl.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if apiServerURL == "" {
+		return nil
+	}
+
+	apiServerURL = strings.TrimSuffix(apiServerURL, "/")
+
+	clusterCopy := managedCluster.DeepCopy()
+	for _, config := range clusterCopy.Spec.ManagedClusterClientConfigs {
+		if config.URL == apiServerURL {
```
zhujian7, line 31: should it also trimSuffix "/" for the `config.URL` before comparing? since the `config.URL` might be configured by users?

pkg/controller/autoimport/apiserverurl.go
```diff
@@ -0,0 +1,103 @@
...
+			return nil
+		}
+	}
+
+	clusterCopy.Spec.ManagedClusterClientConfigs = append(clusterCopy.Spec.ManagedClusterClientConfigs,
+		clusterv1.ClientConfig{
+			URL: apiServerURL,
+		})
+
+	klog.Infof("update the apiServerURL %v to the spec of managedCluster %v.", apiServerURL, clusterCopy.Name)
```
zhujian7, line 41: The web UI only shows the first URL in the list, so I am not sure if append or prepend is proper here, do we want the UI always show the URL in the bootstrap kubeconfig?

pkg/controller/autoimport/apiserverurl.go
zhiweiyin318: done

pkg/controller/autoimport/apiserverurl.go
zhiweiyin318: not required to show which URL.  I think it is ok to always show the first URL.

zhujian7:
LGTM

PR #396, opened 2024-09-24
Add the service-account way of rosa cluster importing.

xuezhaojun:
Issue: https://issues.redhat.com/browse/ACM-10404


xuezhaojun:
/assign @elgnay 

There are several aspects of the overall process for manually importing discovered clusters that need to be confirmed across squads, particularly with the Console and Documentation teams. 

However, it's unlikely that there will be more code changes to the import-controller beyond those in this PR.

PR #404, opened 2024-10-16
Add bootstrap-kubeconfig of current hub as the last one in the list.

xuezhaojun:
Fixes: https://issues.redhat.com/browse/ACM-14994

### DIFF
--- a/pkg/bootstrap/render.go
+++ b/pkg/bootstrap/render.go
@@ -359,6 +359,13 @@ func (b *KlusterletManifestsConfig) Generate(ctx context.Context, clientHolder *
 		if err != nil {
 			return nil, err
 		}
+		// add default bootstrap kubeconfig secret into the list:
+		// TODO: deduplicate the bootstrap kubeconfig secrets @xuezhaojun
+		bootstrapKubeConfigSecrets = append(bootstrapKubeConfigSecrets, BootstrapKubeConfigSecret{
+			Name:       constants.DefaultBootstrapHubKubeConfigSecretName + "-current-hub",
+			KubeConfig: base64.StdEncoding.EncodeToString(b.BootstrapKubeconfig),
+		})
+
 		renderConfig.MultipleHubsEnabled = true
 		renderConfig.BootstrapKubeConfigSecrets = bootstrapKubeConfigSecrets
 	}
--- a/pkg/bootstrap/render_test.go
+++ b/pkg/bootstrap/render_test.go
@@ -653,9 +653,9 @@ func TestKlusterletConfigGenerate(t *testing.T) {
 				},
 			}),
 			validateFunc: func(t *testing.T, objs []runtime.Object) {
-				// 11 objects for klusterlet manifests, 2 objects for bootstrap kubeconfig secrets
-				if len(objs) != 11 {
-					t.Fatalf("Expected 11 objects, but got %d", len(objs))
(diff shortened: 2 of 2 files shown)


pkg/bootstrap/manifests/klusterlet/bootstrap_secret.yaml
```diff
@@ -14,6 +14,19 @@ type: Opaque
 data:
   kubeconfig: "{{ $bootstrapkubeconfigsecret.KubeConfig }}"
 {{end}}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "{{ .DefaultBootstrapKubeConfigSecret.Name }}-current-hub"
```
clyang82, line 21: maybe call it `original-hub`?

xuezhaojun: The `original` imples that is the first, initial hub, but in our implementation, this piece of book-kubeconfig always points to the hub currentlly connectted with.

xuezhaojun:
@clyang82 I realize the implamentation is diff with the story description, because we puting the default kubeconfig in the `klusterlet manifests` not in the `klustereltconfig`. Is this accpetable to global hub's case?

clyang82:
Does that mean you are going to allow configuring only one bootstrap kubeconfig in `kubeConfigSecrets` field?

xuezhaojun:
> Does that mean you are going to allow configuring only one bootstrap kubeconfig in `kubeConfigSecrets` field?

@clyang82 , Yes, allowed in `KlusterletConfig`. We don't have checker on amount of kubeConfigSecrets in import-controller, if you only put one in it, import-controller will populate to 2 in actual `Klusterlet` and delegate it to agent.

pkg/bootstrap/manifests/klusterlet/bootstrap_secret.yaml
```diff
@@ -14,6 +14,19 @@ type: Opaque
 data:
   kubeconfig: "{{ $bootstrapkubeconfigsecret.KubeConfig }}"
 {{end}}
+---
+apiVersion: v1
+kind: Secret
```
qiujian16, line 19: TBH I think we should resolve this in code. I would expect you put it in the BootstrapKubeConfigSecrets list and also remove any duplications in the list

xuezhaojun: Do we mean modify the `KlusterletConfig.BootstrapKubeConfigSecrets` directly here? KlusterConfig needs to bind with managed cluster via annotation, does this mean we need to reconcile every KlusterletConfig and add default bootstrapkubeconfig  of current hub?

qiujian16: yes when klusterletConfig has this part set.

xuezhaojun: We can set it to the klusterletConfig, because the `KubeConfigSecret` in `KlusterletConfig` only has the secret name, and the secret is created in the same namespace with import-controller. 

Not like default bootstrapKubeConfig which data part is generated or decoded from an import-secret. 

So I still put it into the `render.go`, but in a middle stage list -- `bootstrapKubeConfigSecrets`, in the future if we want to add deduplication, we can do it based on this stage.

PR #405, opened 2024-10-20
WIP: Add skip bootstrapkubeconfig option in klusterlet manifests.

xuezhaojun:
With this feature and we can set any bootstrap kubeconfig we want using Manifestwork.

TODO: the annotation should be added in clusterlifecycle-api repo.


xuezhaojun:
Conclusion: Still using Multiplehubs as the only interface to interact with bootstrapkubeconfigs.

PR #414, opened 2024-11-07
Fix typo `regitration` to `registration`

xuezhaojun:
Also updated in backplane-operator: https://github.com/stolostron/backplane-operator/pull/1077

### DIFF
--- a/build/setup-import-controller.sh
+++ b/build/setup-import-controller.sh
@@ -170,7 +170,7 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: managedcluster-import-controller-agent-regitration-client
+  name: managedcluster-import-controller-agent-registration-client
 subjects:
   - kind: ServiceAccount
     name: managed-cluster-import-e2e-agent-registration-sa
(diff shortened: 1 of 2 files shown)


build/setup-import-controller.sh
```diff
@@ -170,7 +170,7 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: managedcluster-import-controller-agent-regitration-client
+  name: managedcluster-import-controller-agent-registration-client
```
zhujian7, line 173: please test the upgrade case, if I remember correctly, the roleRef can not be updated.

zhujian7 (changes requested):

build/setup-import-controller.sh
zhujian7: And for the change in the backplane operator https://github.com/stolostron/backplane-operator/pull/1077/files , why do we only need to update `ClusterRole`, how about the role binding, roleRef there?

xuezhaojun: There is no rolebinding and roleRef for this clusterrole, the clusterrole is created for users who want to use agent-registration, they create their serviceaccount and clusterrolebinding. 

I put this clusterrole in backplane-operator only want to reduce a common clusterrole that users need to create. But users can create the clusterrole by themselves as well.

xuezhaojun: We don't have roleRef need to updated, only this clusterrole.

PR #415, opened 2024-11-07
Add flightctl support.

xuezhaojun:
Related Story: 
* https://issues.redhat.com/browse/ACM-13294
* https://issues.redhat.com/browse/ACM-15480
* https://issues.redhat.com/browse/EDM-194

Related PR:
* https://github.com/stolostron/backplane-operator/pull/1108: need to add a new arg and networkpolicy related permission.

Related Slack:
* Summary: https://redhat-internal.slack.com/archives/C070MTMD4BZ/p1732720170298269

Related Question DOC:
* https://docs.google.com/document/d/1dZqVQ75u-TAacvt0GkXUfAjW1R2mSSAQKlIcBESxNxo/edit?tab=t.0


xuezhaojun:
/assign @qiujian16 

Currently lack of test cases, I'm working on it, but I test it with a ACM 2.13 downstream build,  I think the implementation part is ready to be reviewed.

pkg/controller/flightctl/flightctl.go
```diff
@@ -0,0 +1,186 @@
...
+		return err
+	}
+
+	expectedRepository := &flightctlapiv1.Repository{
+		ApiVersion: "v1alpha1",
+		Kind:       "Repository",
+		Metadata: flightctlapiv1.ObjectMeta{
+			// Note: In the fleets' `httpRef.repository` field, the name is `acm-registration`.
+			// See details in: https://github.com/flightctl/flightctl/blob/main/docs/user/registering-microshift-devices-acm.md
+			Name: ptr("acm-registration"),
```
qiujian16, line 95: nit, there is an existing func pointers.String or something you could use

pkg/controller/flightctl/flightctl.go
```diff
@@ -0,0 +1,186 @@
...
+	flightctlcli "github.com/flightctl/flightctl/lib/cli"
+	"github.com/stolostron/managedcluster-import-controller/pkg/helpers"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	FlightCtlNamespace      = "flightctl"
+	FlightCtlInternalServer = "https://flightctl-api.flightctl.svc.cluster.local:3443"
```
qiujian16, line 20: is this name a constant?

pkg/controller/flightctl/flightctl.go
```diff
@@ -0,0 +1,186 @@
...
+	agentRegistrationServer string
+	cachedToken             string
+	cachedCA                string
+}
+
+var _ FlightCtler = &FlightCtl{}
+
+func (f *FlightCtl) IsFlightCtlEnabled(ctx context.Context) (bool, error) {
+	// If `flightctl` ns is not created, means the flightctl is not enabled.
+	ns := &corev1.Namespace{}
```
qiujian16, line 64: if we already have the server url, why not call server to check if it is enabled?

pkg/controller/flightctl/flightctl.go
```diff
@@ -0,0 +1,186 @@
...
+
+	return true, nil
+}
+
+func (f *FlightCtl) getFlightCtlClientServiceAccountToken(ctx context.Context) (string, error) {
+	if f.cachedToken != "" {
+		return f.cachedToken, nil
+	}
+
+	secret := &corev1.Secret{}
```
qiujian16, line 139: you should use tokenRequest instead of secret. The secret way is not secure and will be disabled by default in the future.

pkg/controller/flightctl/flightctl.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (f *FlightCtl) getAgentRegistrationCA(ctx context.Context) (string, error) {
+	if f.cachedCA != "" {
+		return f.cachedCA, nil
+	}
+
+	configMap := &corev1.ConfigMap{}
+	err := f.clientHolder.RuntimeClient.Get(ctx, types.NamespacedName{
+		Name:      "kube-root-ca.crt",
+		Namespace: "kube-system",
```
qiujian16, line 168: why kube-system? if you want the internal root-ca, it is mounted already in the pod

pkg/controller/flightctl/flightctl.go
```diff
@@ -0,0 +1,186 @@
...
+var files = []string{
+	"manifests/clusterrole.yml",
+	"manifests/clusterrolebinding_agentregistration.yml",
+	"manifests/clusterrolebinding_flightctl.yml",
+	"manifests/serviceaccount.yml",
+	"manifests/serviceaccounttokensecret.yml",
+	"manifests/networkpolicy.yml",
+}
+
+type FlightCtler interface {
```
qiujian16, line 38: why it is an interface? easier for ut?

pkg/controller/controller.go
```diff
@@ -46,7 +48,19 @@ func AddToManager(ctx context.Context,
...
-					[]func(ctx context.Context, csr *certificatesv1.CertificateSigningRequest) (bool, error){})
+					[]func(ctx context.Context, csr *certificatesv1.CertificateSigningRequest) (bool, error){
+						// Case 1: If flightctl is enabled and a csr is from a flightctl device, approve it.
+						func(ctx context.Context, csr *certificatesv1.CertificateSigningRequest) (bool, error) {
+							isFlightCtlEnabled, err := flightctler.IsFlightCtlEnabled(ctx)
+							if err != nil {
+								return false, err
+							}
+							if isFlightCtlEnabled {
+								return flightctler.IsManagedClusterAFlightctlDevice(ctx, csr.Spec.Username)
```
qiujian16, line 59: what if there is an error here? how can the csr reonciler to be triggered again?

pkg/controller/controller.go
```diff
@@ -90,6 +104,14 @@ func AddToManager(ctx context.Context,
 				return nil
 			},
 		},
+		{
+			flightctl.NamespaceControllerName,
```
qiujian16, line 108: I still concern the idea of starting these controllers by default. It brings risk to our normal import flow when flightctl is not enabled.

pkg/controller/flightctl/nscontroller.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func AddNSController(ctx context.Context, mgr manager.Manager, flightctler FlightCtler, clientHolder *helpers.ClientHolder) error {
+	err := ctrl.NewControllerManagedBy(mgr).Named(NamespaceControllerName).
+		Watches(
```
qiujian16, line 94: this will watch all the namespace, could bring scalability issue.

pkg/controller/controller.go
xuezhaojun: If I understanding it right, another way is to use backplane-operator to tell the import-controller if flightctl is enabled or not at the pod start phase.  

I added a question into the doc, we may need to check the progress of flightctl + MCE.

pkg/controller/controller.go
xuezhaojun: We refactor the csr controller before, if it returns err, if will requeue: 

https://github.com/stolostron/managedcluster-import-controller/blob/f65821a693aabfb51b10a2cbf28a5e8db407e3e6/pkg/controller/csr/csr_controller.go#L104C1-L114C3

pkg/controller/flightctl/flightctl.go
xuezhaojun: Updated to tokenRequest way.

pkg/controller/controller.go
xuezhaojun: Need to confirm with the installer side, see if they have plan and buffer to implement a flag we can use.

xuezhaojun: Updated to use a flag to control everything related to flightctl.

pkg/controller/flightctl/nscontroller.go
xuezhaojun: NSController now is removed.

pkg/controller/flightctl/flightctl.go
xuezhaojun: Update to a internal interface, change flightctl to flightctlManager(a structure)

xuezhaojun:
@qiujian16 , the PR is updated, pleaset take another look when you have time! Thanks!

pkg/controller/flightctl/flightctl.go
```diff
@@ -0,0 +1,254 @@
...
+		CreateToken(ctx, "flightctl-client", tokenRequest, metav1.CreateOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to create token: %v", err)
+	}
+
+	f.cachedToken = tokenResponse.Status.Token
+	return f.cachedToken, nil
+}
+
+func (f *FlightCtlManager) getAgentRegistrationCA() (string, error) {
```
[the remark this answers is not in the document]
xuezhaojun: TODO added.

pkg/controller/flightctl/flightctl.go
```diff
@@ -0,0 +1,254 @@
...
+	if f.cachedCA != "" {
+		return f.cachedCA, nil
+	}
+
+	caData, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")
```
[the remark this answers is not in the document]
xuezhaojun: Remove the cachedCA then we can always get the rotated, latest CA.

PR #416, opened 2024-11-07
Refactor: controller `AddToManager`

xuezhaojun:
Simply `AddToManager`, and:
1. When we call `Add` function, only need to input the varibles the controller actually needs. For example, csrContoller doesn't need `informer` and `recoder`, now it doesn't need put them in the `Add`'s `inputs`.
2. In the future, if we want to add customized varibles to specific controller, for example, if we want to add `flightctl flag` to csr controller, without affect other controllers.
3. Now hosted controller can handle like other controllers. Unified the code style.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -27,28 +27,8 @@ import (
 	"github.com/stolostron/managedcluster-import-controller/pkg/source"
 
 	kevents "k8s.io/client-go/tools/events"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
-
-var log = logf.Log.WithName("controllers")
-
-type AddToManagerFunc func(
-	context.Context, manager.Manager, *helpers.ClientHolder, *source.InformerHolder, kevents.EventRecorder,
-) (string, error)
-
-// AddToManagerFuncs is a list of functions to add all controllers to the manager
-var AddToManagerFuncs = []AddToManagerFunc{
-	csr.Add,
-	managedcluster.Add,
-	importconfig.Add,
-	manifestwork.Add,
-	selfmanagedcluster.Add,
-	autoimport.Add,
-	clusterdeployment.Add,
-	clusternamespacedeletion.Add,
-	importstatus.Add,
-}
 
 // AddToManager adds all controllers to the manager
 func AddToManager(ctx context.Context,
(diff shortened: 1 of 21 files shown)


pkg/controller/controller.go
```diff
@@ -27,51 +27,42 @@ import (
...
-		if err != nil {
-			return err
-		}
 
-		log.Info(fmt.Sprintf("Add controller %s to manager", name))
+	controllers := []struct {
+		name string
+		add  func() error
+	}{
+		{csr.ControllerName, func() error { return csr.Add(ctx, manager, clientHolder) }},
```
zhujian7, line 44: Nit: can we format these controllers to improve code readability like:
```
	controllers := []struct {
		name string
		add  func() error
	}{
		{
			csr.ControllerName,
			func() error { return csr.Add(ctx, manager, clientHolder) },
		},
		{
			managedcluster.ControllerName,
			func() error { return managedcluster.Add(ctx, manager, clientHolder, mcRecorder) },
		},
		{
			importconfig.ControllerName,
			func() error { return importconfig.Add(ctx, manager, clientHolder, informerHolder) },
		},
		{
			manifestwork.ControllerName,
			func() error { return manifestwork.Add(ctx, manager, clientHolder, informerHolder, mcRecorder) },
		},
		{
			selfmanagedcluster.ControllerName,
			func() error { return selfmanagedcluster.Add(ctx, manager, clientHolder, informerHolder, mcRecorder) },
		},
		{
			autoimport.ControllerName,
			func() error { return autoimport.Add(ctx, manager, clientHolder, informerHolder, mcRecorder) },
		},
		{
			clusterdeployment.ControllerName,
			func() error { return clusterdeployment.Add(ctx, manager, clientHolder, informerHolder, mcRecorder) },
		},
		{
			clusternamespacedeletion.ControllerName,
			func() error { return clusternamespacedeletion.Add(ctx, manager, clientHolder) },
		},
		{
			importstatus.ControllerName,
			func() error { return importstatus.Add(ctx, manager, clientHolder, informerHolder, mcRecorder) },
		},
		{
			hosted.ControllerName,
			func() error {
				if features.DefaultMutableFeatureGate.Enabled(features.KlusterletHostedMode) {
					return hosted.Add(ctx, manager, clientHolder, informerHolder, mcRecorder)
				}
				return nil
			}},
	}
```

zhujian7:
/hold
for @zhiweiyin318 to review

pkg/controller/controller.go
xuezhaojun: Good point, updated.

pkg/controller/controller.go
```diff
@@ -27,51 +27,42 @@ import (
...
 	informerHolder *source.InformerHolder,
 	mcRecorder kevents.EventRecorder) error {
-	for _, addFunc := range AddToManagerFuncs {
-		name, err := addFunc(ctx, manager, clientHolder, informerHolder, mcRecorder)
-		if err != nil {
-			return err
-		}
 
-		log.Info(fmt.Sprintf("Add controller %s to manager", name))
+	controllers := []struct {
```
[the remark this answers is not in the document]
xuezhaojun: I'm not sure we need another layer of abstruction, it seems increase the overall compliexity, however it's a good idea to rename the array for better readbiltiy. 

Updated.

PR #419, opened 2024-11-14
Refactor csr controller to prepare extend more checks(for example, check if a csr is from flightctl device)

xuezhaojun:


pkg/controller/csr/manager.go
```diff
@@ -45,6 +47,22 @@ func Add(ctx context.Context,
...
+					err := clientHolder.RuntimeClient.Get(ctx, types.NamespacedName{Name: clusterName}, &cluster)
+					if errors.IsNotFound(err) {
+						return false, nil
+					}
+					if err != nil {
+						return false, err
+					}
+					return true, nil
+				},
+			},
```
xuezhaojun, line 65: More cases with clear description will be added here.

PR #420, opened 2024-11-15
Append the current kubeconfig name into registrationConfiguration.bootstrapKubeConfigs.localSecretsConfig.kubeConfigSecrets

clyang82:
In this PR https://github.com/stolostron/managedcluster-import-controller/pull/404, we put the secret into import.yaml so that the secret can be created in `open-cluster-management-agent` namespace. 
This PR is to put the `-current-hub` secret name into registrationConfiguration.bootstrapKubeConfigs.localSecretsConfig.kubeConfigSecrets so that if the user only provide one secret, we will append the original one. Then it can meet our desired behaviour.

ref: https://issues.redhat.com/browse/ACM-14994

### DIFF
--- a/pkg/bootstrap/render.go
+++ b/pkg/bootstrap/render.go
@@ -353,6 +353,10 @@ func (b *KlusterletManifestsConfig) Generate(ctx context.Context, clientHolder *
 				Mode:    operatorv1.FeatureGateModeTypeEnable,
 			})
 		registrationConfiguration.BootstrapKubeConfigs = b.klusterletconfig.Spec.BootstrapKubeConfigs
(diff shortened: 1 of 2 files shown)


xuezhaojun:
Good catch, you're right, both fields affect in render process:
* BootstrapKubeConfigSecrets used in: https://github.com/stolostron/managedcluster-import-controller/blob/fcc928c70cc80a2aff44cd7d620c138dc5fd932a/pkg/bootstrap/manifests/klusterlet/bootstrap_secret.yaml#L2
* RegistrationConfiguration used in: https://github.com/stolostron/managedcluster-import-controller/blob/fcc928c70cc80a2aff44cd7d620c138dc5fd932a/pkg/bootstrap/manifests/klusterlet/klusterlet.yaml#L17


/lgtm

PR #421, opened 2024-11-15
Refactor csr controller to prepare extend more checks(for example, check if a csr is from flightctl device)

xuezhaojun:
…eck if a csr is from flightctl device)

### DIFF
--- a/pkg/controller/csr/csr_controller.go
+++ b/pkg/controller/csr/csr_controller.go
@@ -9,13 +9,13 @@ import (
 
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/stolostron/managedcluster-import-controller/pkg/helpers"
-	clusterv1 "open-cluster-management.io/api/cluster/v1"
 
 	certificatesv1 "k8s.io/api/certificates/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
 
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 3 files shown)


pkg/controller/csr/csr_controller.go
```diff
@@ -90,15 +89,21 @@ func (r *ReconcileCSR) Reconcile(ctx context.Context, request reconcile.Request)
 		return reconcile.Result{}, nil
 	}
 
-	clusterName := getClusterName(csr)
-	cluster := clusterv1.ManagedCluster{}
-	err = r.clientHolder.RuntimeClient.Get(ctx, types.NamespacedName{Name: clusterName}, &cluster)
-	if errors.IsNotFound(err) {
-		// no managed cluster, do nothing.
-		return reconcile.Result{}, nil
+	// Run checks one by one, if any check returns true, approve the CSR
```
zhujian7, line 92: Intuitively, all checks should be passed before approval?

pkg/controller/csr/manager.go
```diff
@@ -45,6 +47,21 @@ func Add(ctx context.Context,
 		Complete(&ReconcileCSR{
 			clientHolder: clientHolder,
 			recorder:     helpers.NewEventRecorder(clientHolder.KubeClient, ControllerName),
+			checks: []func(ctx context.Context, csr *certificatesv1.CertificateSigningRequest) (bool, error){
+				// Case 1: if a CSR comes from a managed cluster that already exists, approve it
```
zhujian7, line 51: based on the current checks logic, as long as the managedcluster exists, it should be approved even other checks are not passed?

pkg/controller/csr/csr_controller.go
```diff
@@ -90,15 +89,21 @@ func (r *ReconcileCSR) Reconcile(ctx context.Context, request reconcile.Request)
 		return reconcile.Result{}, nil
 	}
 
-	clusterName := getClusterName(csr)
-	cluster := clusterv1.ManagedCluster{}
-	err = r.clientHolder.RuntimeClient.Get(ctx, types.NamespacedName{Name: clusterName}, &cluster)
-	if errors.IsNotFound(err) {
-		// no managed cluster, do nothing.
```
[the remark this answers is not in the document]
zhujian7: I am not sure if this should be a check, I think this seems like a common requirement before all other checks. it is like a prerequisite.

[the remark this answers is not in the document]
xuezhaojun: Yes, in cases like agent-registration and flightctl, the managed cluster is created by agent-side, the managed cluster is not promised to be created when csr is created.

xuezhaojun:
@zhujian7  You're right, this csr controller is meant to only handle csr from managed cluster, this line of code make sure only csr from managed cluster get handled, the agent-registration csr won't be watched in this controlelr.

https://github.com/stolostron/managedcluster-import-controller/blob/b1ccdeaa020f31ac73023f28506967782c442eea/pkg/controller/csr/manager.go#L39

Thanks for helping me think it through! 

/close

pkg/controller/csr/csr_controller.go
```diff
@@ -53,17 +53,38 @@ func validUsername(csr *certificatesv1.CertificateSigningRequest, clusterName st
...
-func csrPredicate(csr *certificatesv1.CertificateSigningRequest) bool {
+// isValidUnapprovedBootstrapCSR checks if the CSR:
+// 1. Has a non-empty cluster name label
+// 2. Has not been approved or denied
+func isValidUnapprovedBootstrapCSR(csr *certificatesv1.CertificateSigningRequest) bool {
 	clusterName := getClusterName(csr)
 	return clusterName != "" &&
-		getApprovalType(csr) == "" &&
-		validUsername(csr, clusterName)
+		getApprovalType(csr) == ""
```
xuezhaojun, line 62: @zhujian7 The `validUsername` function limit watched csrs exclude agent-registration type. So I remove it from watches function.

pkg/controller/csr/csr_controller.go
```diff
@@ -53,17 +53,38 @@ func validUsername(csr *certificatesv1.CertificateSigningRequest, clusterName st
...
+}
+
+// approveExistingManagedClusterCSR checks if the CSR is from an existing managed cluster
+func approveExistingManagedClusterCSR(ctx context.Context, csr *certificatesv1.CertificateSigningRequest,
+	clientHolder *helpers.ClientHolder) (bool, error) {
+	if !validUsername(csr, getClusterName(csr)) {
```
xuezhaojun, line 68: @zhujian7 The `validUsername` is put into the default approveCondition, so old logic still work.

pkg/controller/csr/csr_controller.go
```diff
@@ -90,15 +111,21 @@ func (r *ReconcileCSR) Reconcile(ctx context.Context, request reconcile.Request)
...
 
-	clusterName := getClusterName(csr)
-	cluster := clusterv1.ManagedCluster{}
-	err = r.clientHolder.RuntimeClient.Get(ctx, types.NamespacedName{Name: clusterName}, &cluster)
-	if errors.IsNotFound(err) {
-		// no managed cluster, do nothing.
-		return reconcile.Result{}, nil
+	// Check if any approval condition matches
+	shouldApprove := false
+	for _, condition := range r.approvalConditions {
```
xuezhaojun, line 116: @zhujian7 
The `checks` is renamed to `appprovalConditions` to avoid misunderstanding.

PR #425, opened 2024-11-21
use internal endpoint for local-cluster

elgnay:
The related issue: https://issues.redhat.com/browse/ACM-15486


pkg/controller/importconfig/cluster_info.go
```diff
@@ -144,3 +109,183 @@ func validateToken(token string, creation, expiration []byte) bool {
...
+
+	if importSecret == nil {
+		klog.Infof("the import secret is missing for the managed cluster %s", managedCluster.Name)
+	}
+
+	// check if the bootstrap kubeconfig and token in the import secret are still valid
+	if kubeconfigData := extractBootstrapKubeConfigDataFromImportSecret(importSecret); len(kubeconfigData) > 0 {
+		kubeAPIServer, proxyURL, ca, caData, tokenString, ctxClusterName, err := parseKubeConfigData(kubeconfigData)
+		if err != nil {
+			klog.Errorf("failed to parse the bootstrap hub kubeconfig in the import.yaml. Recreation is required: %v", err)
```
zhujian7, line 145: why do not we return here?

pkg/controller/importconfig/cluster_info.go
```diff
@@ -144,3 +109,183 @@ func validateToken(token string, creation, expiration []byte) bool {
 	lifetime := time.Until(expirationTime)
 	return lifetime > refreshThreshold
 }
+
+func buildBootstrapKubeconfigData(ctx context.Context, clientHolder *helpers.ClientHolder,
+	managedCluster *clusterv1.ManagedCluster,
+	klusterletConfig *klusterletconfigv1alpha1.KlusterletConfig) ([]byte, []byte, []byte, error) {
+	var bootstrapKubeconfigData, tokenData, tokenCreation, tokenExpiration []byte
```
zhujian7, line 116: I thought the internal endpoint kubeconfig has a fixed format, so we can return it directly and keep the other logic unchanged. like:
```
if isSelfManaged(managedCluster) {
  return internal-kubeconfig-bootstrap-config-data,....
}

...
```

elgnay: The internal endpoint will be used only when:
- The managed cluster is the local-clsuter;
- And there is no custom config of kube-apiserver url and CA;

So the user is still able to change the kube-apiserver url or CA of a self managed cluster by using the klusterletconfig API.

elgnay: This PR also refactors the logic to create the bootstrap kubeconfig. It becomes simpler and uses less API calls to the apiserver now.

pkg/bootstrap/bootstrapkubeconfig.go
```diff
@@ -87,7 +84,54 @@ func CreateBootstrapKubeConfig(ctx context.Context, clientHolder *helpers.Client
...
 
-const bootstrapSASuffix = "bootstrap-sa"
+func GetKubeAPIServerConfig(ctx context.Context, clientHolder *helpers.ClientHolder, ns string,
+	klusterletConfig *klusterletconfigv1alpha1.KlusterletConfig, selfManaged bool) (url, proxy, ca string, caData []byte, err error) {
+	proxy, _ = GetProxySettings(klusterletConfig)
+	// use internal endpoint by default for the self managed cluster (local-cluster)
+	// https proxy with additional CA data is not supported becuase the CAs used in
+	// this situation is loaded from a file in the pod
+	if selfManaged && !hasCustomServerURLOrCA(klusterletConfig) {
+		return apiServerInternalEndpoint, proxy, apiServerInternalEndpointCA, caData, err
```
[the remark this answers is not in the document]
elgnay: You are right. If the internal endpoint is used, we might need to ignore the proxy settings.

pkg/controller/importconfig/cluster_info.go
elgnay: It will result in an infinite loop if an error is returned when the existing kubeconfig is invalid. We'd better to log the error and regenerate the bootstrap kubeconfig in this case.

elgnay:
> @elgnay Is it possible we can sperate logic change and refactor into 2 or more PRs?

@xuezhaojun Yes. Here is the new PR for the code refactor: https://github.com/stolostron/managedcluster-import-controller/pull/427. Please take a look.

pkg/bootstrap/bootstrapkubeconfig.go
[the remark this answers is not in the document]
elgnay: Code updated to ignore the proxy setting.

pkg/controller/importconfig/cluster_info.go
elgnay: Changed the log level to `Info`.

elgnay:
E2e test cases added.

pkg/bootstrap/bootstrapkubeconfig.go
```diff
@@ -39,19 +39,27 @@ import (
 )
 
 const (
-	bootstrapSASuffix = "bootstrap-sa"
+	bootstrapSASuffix           = "bootstrap-sa"
+	apiServerInternalEndpoint   = "https://kubernetes.default.svc:443"
```
zhujian7, line 43: I am thinking if we deploy the MCE on a ROSA-HCP(hosted control plane), this seems not correct, this is the hosting cluster API server address, not the hcp API server address, right?
we need to test this case before merging.

PR #428, opened 2024-11-22
Refactor: move common functions out of csr controller, into `constants` or `helpers`.

xuezhaojun:
There are mainly **2** changes in this PR:

**First**, the `GetClusterName` is moved from `csr-controller` to `helpers` pkg, along with some const values move to `constant` pkg.

**Second**, the `Add` function added a `extraApprovalConditions` array, to allow the `csr controller` to accept and use customized approval conditions.

---

**What I want to achieve**, is in the future, we can have more approvalConditions defined and implementated in their our pkg(for example, flightctl approve stays in the `flightctl` pkg with other flight logic code),  when we add new approval condition, the `csr controller` don't need to change much.

### DIFF
--- a/pkg/controller/csr/csr_controller.go
+++ b/pkg/controller/csr/csr_controller.go
@@ -23,19 +23,9 @@ import (
 
 const (
 	userNameSignature = "system:serviceaccount:%s:%s-bootstrap-sa"
-	clusterLabel      = "open-cluster-management.io/cluster-name"
 )
 
 var log = logf.Log.WithName("controller_csr")
-
-func getClusterName(csr *certificatesv1.CertificateSigningRequest) (clusterName string) {
-	for label, v := range csr.GetObjectMeta().GetLabels() {
-		if label == clusterLabel {
-			clusterName = v
-		}
-	}
-	return clusterName
-}
 
 func getApprovalType(csr *certificatesv1.CertificateSigningRequest) string {
 	if csr.Status.Conditions == nil {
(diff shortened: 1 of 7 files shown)


pkg/controller/csr/csr_controller.go
```diff
@@ -23,20 +23,10 @@ import (
 
 const (
 	userNameSignature = "system:serviceaccount:%s:%s-bootstrap-sa"
-	clusterLabel      = "open-cluster-management.io/cluster-name"
 )
 
 var log = logf.Log.WithName("controller_csr")
 
-func getClusterName(csr *certificatesv1.CertificateSigningRequest) (clusterName string) {
```
xuezhaojun, line 31: This is the first change, move to helpers pkg

pkg/controller/csr/manager.go
```diff
@@ -17,15 +17,15 @@ import (
...
+	ControllerName = "csr-controller"
 )
 
 // Add creates a new CSR Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(ctx context.Context,
 	mgr manager.Manager,
-	clientHolder *helpers.ClientHolder) error {
+	clientHolder *helpers.ClientHolder,
+	extraApprovalConditions []func(ctx context.Context, csr *certificatesv1.CertificateSigningRequest) (bool, error)) error {
```
xuezhaojun, line 28: Here is the second change, now the Add function accept extraApprovalConditions.

PR #452, opened 2025-01-04
WIP: Sync ClusterDeployment Unreachable condition to managedcluster for UI/Console to leverage.

xuezhaojun:
Jira: https://issues.redhat.com/browse/ACM-16340

The source condition is from: https://github.com/openshift/hive/blob/a428d219e9fb59937d9b7e847f8d921ffb7b59a8/apis/hive/v1/clusterdeployment_types.go#L426

The controller in hive maintain this condition: https://github.com/openshift/hive/blob/a428d219e9fb59937d9b7e847f8d921ffb7b59a8/pkg/controller/unreachable/unreachable_controller.go#L277

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -14,7 +14,9 @@ import (
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -158,6 +160,11 @@ func (r *ReconcileClusterDeployment) Reconcile(
 		}
 	}
 
+	// sync the Unreachable condition to ManagedCluster
+	if err := r.syncClusterReacableCondition(ctx, managedCluster, clusterDeployment); err != nil {
+		return reconcile.Result{}, err
+	}
+
 	return result, iErr
 }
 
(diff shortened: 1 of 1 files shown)


xuezhaojun:
/assign @qiujian16
/assign @elgnay 

Request for the review to confirm the direction of this solution is right.

elgnay:
I have two comments in general:
- Perhaps the `ocm-controller` is a better component to run this new controller;
- `Unreachable` condition from Hive might include information beyond our expectation. What we expected is to have a condition to indicate if the admin-kubeconfig of the managed cluster is invalid or not.

xuezhaojun:
> I have two comments in general:
> 
> * Perhaps the `ocm-controller` is a better component to run this new controller;
> * `Unreachable` condition from Hive might include information beyond our expectation. What we expected is to have a condition to indicate if the admin-kubeconfig of the managed cluster is invalid or not.

Thanks for the review!

---

I checked the `ocm-controller` controllers, it seems it doesn't have a specific controller that handles the `clusterdeployment` resource.  

The most close one is under the `/clusterset` folder: 
* https://github.com/stolostron/multicloud-operators-foundation/blob/647e6993139536191a417f4f6d9e7a2b6074bbf3/pkg/controllers/clusterset/clusterdeployment/clusterdeployment_controller.go 

Compare with the import-controller, the path of the controller only has a the `/clusterdeployment`:  
* https://github.com/stolostron/managedcluster-import-controller/blob/main/pkg/controller/clusterdeployment/clusterdeployment_controller.go

---

Because the CA update is happens from the agent side, so to check the valid of the admin-kubeconfig, we need to implement the `polling` strategy on the hub side.

In hive, they use a [DelayingReconciler](https://github.com/openshift/hive/blob/a428d219e9fb59937d9b7e847f8d921ffb7b59a8/pkg/controller/unreachable/unreachable_controller.go#L92C44-L92C65) to check valid status every [1000 seconds](https://github.com/openshift/hive/blob/a428d219e9fb59937d9b7e847f8d921ffb7b59a8/pkg/controller/utils/delayingreconciler.go#L11).

And hive also **ensure the performance at scale** by using the [MaxConcurrentReconciles](https://github.com/openshift/hive/blob/a428d219e9fb59937d9b7e847f8d921ffb7b59a8/pkg/controller/unreachable/unreachable_controller.go#L93-L94) and [some hard code retry logic e.g. maxUnreachableDuration](https://github.com/openshift/hive/blob/a428d219e9fb59937d9b7e847f8d921ffb7b59a8/pkg/controller/unreachable/unreachable_controller.go#L53)

If we want to do the same valid check, we need to introduce these complixities to our repo and consider the performance issue as well.  

So to me, I prefer, for the first step, we only do syncing the condition to managed cluster to let console have something to display. If customer like the feature and use it a lot, then we put effort to enhance it.

PR #481, opened 2025-02-10
Change to the same way of linting as ocm repo, and upgrade golint-ci version and configuration.

xuezhaojun:
golint get fail in the old way: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/stolostron_managedcluster-import-controller/475/pull-ci-stolostron-managedcluster-import-controller-main-check/1887846831249952768


golangci.yml
```diff
@@ -188,7 +181,26 @@ linters-settings:
...
+  gci:
+    # Section configuration to compare against.
+    # Section names are case-insensitive and may contain parameters in ().
+    # The default order of sections is `standard > default > custom > blank > dot`,
+    # If `custom-order` is `true`, it follows the order of `sections` option.
+    # Default: ["standard", "default"]
+    sections:
+      - standard # Standard section: captures all standard packages.
+      - default # Default section: contains all imports that could not be matched to another section type.
+      - prefix(open-cluster-management.io) # Custom section: groups all imports with the specified Prefix.
```
zhujian7, line 193: The upstream ocm has a very strict package import order check:
1. standard packages
2. 3rd party packages
3. ocm org packages
4. ocm local repo packages

Examples: https://github.com/open-cluster-management-io/ocm/blob/main/pkg/operator/operators/klusterlet/controllers/klusterletcontroller/klusterlet_cleanup_controller.go#L4-L31

I don't know if this kind of strict rule is welcomed for the import controller repo.

xuezhaojun: That's a good point, I didn't add gci this time.

cmd/manager/main.go
```diff
@@ -124,7 +124,11 @@ func main() {
 	pflag.Parse()
 
 	logs.InitLogs()
-	defer logs.FlushLogs()
+	exitCode := 0
```
xuezhaojun, line 127: exitAfterDefer: os.Exit will exit, and `defer logs.FlushLogs()` will not run (gocritic)

.golangci.yml
```diff
@@ -214,8 +127,15 @@ issues:
   # Default value for this option is true.
   exclude-use-default: true
 
-  # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
-  max-per-linter: 0
```
zhujian7, line 218: why remove this?

Makefile
```diff
@@ -51,7 +51,8 @@ check-copyright:
 
 .PHONY: lint
 lint:
-	build/run-lint-check.sh
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.63.4
```
zhujian7, line 54: is it possible to install the golint to the current project sub dir, so it will affect the global linter?

.golangci.yml
xuezhaojun: The field is not supportted by the latest golangci-lint, it will report error.

Makefile
xuezhaojun: Good point! Updated to put golint into `_output` dir.

.golangci.yml
xuezhaojun: It's an dup of `max-issues-per-linter: 0` set above.

xuezhaojun:
The EC check occasionally passes and sometimes fails. I'm unsure whether we need to address it.

zhujian7:
> The EC check occasionally passes and sometimes fails. I'm unsure whether we need to address it or overwrite it.

it seems there are 2 new cves:
```
✕ [Violation] cve.cve_blockers
  ImageRef: quay.io/redhat-user-workloads/crt-redhat-acm-tenant/managedcluster-import-controller-addon-mce-28@sha256:2041ac766ce359bddcdc1f71b94322f71d05a50b93f63c78857ac790103367b1
  Reason: Found 1 CVE vulnerabilities of high security level
  Title: Blocking CVE check
  Description: The SLSA Provenance attestation for the image is inspected to ensure CVEs that have a known fix and meet a certain
  security level have not been detected. If detected, this policy rule will fail. By default, only CVEs of critical and high
  security level cause a failure. This is configurable by the rule data key `restrict_cve_security_levels`. The available levels
  are critical, high, medium, low, and unknown. In addition to that leeway can be granted per severity using the `cve_leeway` rule
  data key containing days of allowed leeway, measured as time between found vulnerability's public disclosure date and current
  effective time, per severity level. To exclude this rule add "cve.cve_blockers:high" to the `exclude` section of the policy
  configuration.
  Solution: Make sure to address any CVE's related to the image. The CVEs are detected by the task that runs a Clair scan and
  emits a result named `SCAN_OUTPUT`.

✕ [Violation] cve.cve_blockers
  ImageRef: quay.io/redhat-user-workloads/crt-redhat-acm-tenant/managedcluster-import-controller-addon-mce-28@sha256:15633db1ce2d463f51a4af5fd6e7f066dc39da1515d41704408475b36e74018e
  Reason: Found 2 CVE vulnerabilities of high security level
  Title: Blocking CVE check
  Description: The SLSA Provenance attestation for the image is inspected to ensure CVEs that have a known fix and meet a certain
  security level have not been detected. If detected, this policy rule will fail. By default, only CVEs of critical and high
  security level cause a failure. This is configurable by the rule data key `restrict_cve_security_levels`. The available levels
  are critical, high, medium, low, and unknown. In addition to that leeway can be granted per severity using the `cve_leeway` rule
  data key containing days of allowed leeway, measured as time between found vulnerability's public disclosure date and current
  effective time, per severity level. To exclude this rule add "cve.cve_blockers:high" to the `exclude` section of the policy
  configuration.
  Solution: Make sure to address any CVE's related to the image. The CVEs are detected by the task that runs a Clair scan and
  emits a result named `SCAN_OUTPUT`.
```

PR #485, opened 2025-02-12
Fixes: flightctl is installed in the mch namespace; Also avoid pod crash if flightctl resources apply fails.

xuezhaojun:
Fixes: https://issues.redhat.com/browse/ACM-17801


cmd/manager/main.go
```diff
@@ -108,9 +108,11 @@ func main() {
 
 	var clusterIngressDomain string
 	var enableFlightCtl bool = false
-
+	var flightctlServer string
 	pflag.StringVar(&clusterIngressDomain, "cluster-ingress-domain", "", "the ingress domain of the cluster")
 	pflag.BoolVar(&enableFlightCtl, "enable-flightctl", false, "enable flightctl")
+	pflag.StringVar(&flightctlServer, "flightctl-server",
+		"https://flightctl-api.open-cluster-management.svc.cluster.local:3443", "the server address of the flightctl")
```
qiujian16, line 115: no need to add .cluster.local

pkg/controller/flightctl/flightctl.go
```diff
@@ -54,45 +50,92 @@ type FlightCtlManager struct {
...
+	flightctlServer         string
 	agentRegistrationServer string
 
 	cachedToken string
 }
 
-func (f *FlightCtlManager) ApplyResources(ctx context.Context) error {
-	var err error
+// StartReconcileFlightCtlResources starts a loop to continuously reconcile FlightCtl resources
+func (f *FlightCtlManager) StartReconcileFlightCtlResources(ctx context.Context) {
```
xuezhaojun, line 60: @qiujian16 @elgnay Change to reconcile the flightctl resources every 5 mins. No crash when error happens.

pkg/controller/flightctl/flightctl.go
```diff
@@ -54,45 +50,92 @@ type FlightCtlManager struct {
...
 }
 
-func (f *FlightCtlManager) applyKuberentesResources(ctx context.Context) error {
-	var err error
+// If the flightctl server is not set, then:
+// 1. list all services in the cluster scope, find the one with name "flightctl-api", and set the server to the cluster IP of the service.
+// 2. if not found, return an error.
+// TODO: @xuezhaojun, this is a temporary solution, because in this release, `flightctl-server` is not pass by installer, will remove
+// this ensureFlightCtlServer() after `flightctl-server` is passed by installer.
+func (f *FlightCtlManager) ensureFlightCtlServer() error {
```
xuezhaojun, line 109: @qiujian16 @elgnay Add the `ensureFlightCtlServer` function to handle customized ACM namespace cases in this release, we have a [follow-up story for installer](https://issues.redhat.com/browse/ACM-17903), but since the release is feature freeze, it would be better to scehdule into next release.

pkg/controller/flightctl/flightctl.go
```diff
@@ -54,45 +50,92 @@ type FlightCtlManager struct {
...
+	if f.flightctlServer != "" {
+		return nil
+	}
 
-	// Check if the FlightCtl namespace exists
-	_, err = f.clientHolder.KubeClient.CoreV1().Namespaces().Get(ctx, FlightCtlNamespace, metav1.GetOptions{})
+	services, err := f.clientHolder.KubeClient.CoreV1().Services("").List(context.TODO(), metav1.ListOptions{})
```
qiujian16, line 114: this could be quite expensive to list all services. Why call list instead of using lister?

pkg/controller/flightctl/flightctl.go
```diff
@@ -54,45 +50,92 @@ type FlightCtlManager struct {
...
+		// Record successful sync
+		f.recorder.Event("ResourcesSynced", "Successfully synced FlightCtl resources")
 	}
 
-	return nil
+	// Run immediately for the first time
+	applyFunc()
+
+	// Then run every 5 minutes
+	for {
```
qiujian16, line 94: There is wait.PollImmediateUntil so you can stop doing it after these are created.

pkg/controller/flightctl/flightctl.go
```diff
@@ -54,45 +50,92 @@ type FlightCtlManager struct {
...
+			f.flightctlServer = fmt.Sprintf("https://flightctl-api.%s.svc:3443", service.Namespace)
+			f.flightctlClient = &flightctlClientImpl{flightctlServer: f.flightctlServer}
+			return nil
+		}
+	}
+
+	return fmt.Errorf("flightctl-api service not found")
+}
+
+// filghtctl is installed in the mch namespace, so no need to check if the namespace exists.
```
qiujian16, line 130: this is not correct comment.

pkg/controller/flightctl/flightctl.go
```diff
@@ -182,7 +225,7 @@ func (f *FlightCtlManager) getFlightCtlClientServiceAccountToken(ctx context.Con
 	}
 
 	// Get the token using TokenRequest API
-	tokenResponse, err := f.clientHolder.KubeClient.CoreV1().ServiceAccounts(FlightCtlNamespace).
+	tokenResponse, err := f.clientHolder.KubeClient.CoreV1().ServiceAccounts(os.Getenv("POD_NAMESPACE")).
```
qiujian16, line 228: I am not sure why we do this. If it is the same namespace, can we directly use the import controller's sa and give it sufficient permission to access flightctl. So that we do not need to call api to get token, but read from file

xuezhaojun: The token will be delivered to flightctl-agent to access agentresgistraion endpoint of the import-controller, so we create token with specific 7 days duration to avoid credentials.

We used to bind 2 clusterroles with the serviceaccount `flightctl-client`, but now it seems a bug, I will seperate the clusterrolebinding.

pkg/controller/flightctl/flightctl.go
xuezhaojun: You're right, removed.

pkg/controller/flightctl/flightctl.go
xuezhaojun: Sure, we can change to Lister.

pkg/controller/flightctl/flightctl.go
xuezhaojun: I did a second think, we need to keep it reconcile to keep the token fresh in flightctl Repository.

xuezhaojun: Change to "after resources applied, keep update Repository every day"

pkg/controller/flightctl/flightctl.go
xuezhaojun: The clusterrolebinding sperated, now it's:
* `flightctl-agent-registration` <-> `flightctl-agent-registration`:  used in the flightctl Repository, will be delivered to the flightctl-agent in managed cluster side.
* `managedcluster-import-controller-v2` <-> `flightctl-client`: used for import-controller to access the flightctl-api service on the hub side.

xuezhaojun:
/hold

Need to test at least once with downstream build.

xuezhaojun:
/unhold

tested

qiujian16:
/lgtm
/hold 
in case @elgnay would like to take a look also.

PR #506, opened 2025-02-24
fix addon cannot be deleted after cluster is gone

zhiweiyin318:


pkg/controller/managedcluster/managedcluster_controller_test.go
```diff
@@ -242,6 +242,30 @@ func TestReconcile(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "managed clusters is deleting, no addon in the ns",
```
zhujian7, line 246: will this cover the changes? line 161.
```
	if len(addons.Items) != 0 {
		// for safety force delete remained addon again here.
		if err = r.deleteManagedClusterAddon(ctx, managedCluster); err != nil {
		}

		return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
	}
```

zhiweiyin318: covered in another case

PR #603, opened 2025-04-01
refactor klusterlet manifests using klusterlet helm chart

zhiweiyin318:
story: https://issues.redhat.com/browse/ACM-18995


pkg/bootstrap/render.go
```diff
@@ -19,214 +19,183 @@ import (
...
+		ReplicaCount: 1,
+		Images: chart.ImagesConfig{
+			ImagePullPolicy: corev1.PullIfNotPresent,
+			ImageCredentials: chart.ImageCredentials{
+				// set true by default
+				CreateImageCredentials: true,
+			},
+		},
+		// the release ns and the agent ns are same, so disable to generate ns.
+		CreateNamespace: true,
```
zhujian7, line 84: the comment said "disable to generate ns", but the coce is true, please confirm.

pkg/bootstrap/render.go
```diff
@@ -19,214 +19,183 @@ import (
...
 
-	generateImagePullSecret bool // by default is true, in hosted mode, it will be set false
-}
+			RunAsNonRoot:             &runAsNonRoot,
+			ReadOnlyRootFilesystem:   &readOnlyRootFilesystem,
+			AllowPrivilegeEscalation: &allowPrivilegeEscalation,
+		},
+		Resources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("2Gi"),
```
zhujian7, line 100: in the last release, @haoqing0110 removed the limits for all components, so I am not sure if this should be removed.

pkg/bootstrap/render.go
```diff
@@ -19,214 +19,183 @@ import (
...
-func NewKlusterletManifestsConfig(installMode operatorv1.InstallMode,
-	clusterName string, bootstrapKubeconfig []byte) *KlusterletManifestsConfig {
-	return &KlusterletManifestsConfig{
-		InstallMode:             installMode,
-		ClusterName:             clusterName,
-		BootstrapKubeconfig:     bootstrapKubeconfig,
-		generateImagePullSecret: true,
+	if chartConfig.Klusterlet.Mode == operatorv1.InstallModeHosted ||
+		chartConfig.Klusterlet.Mode == operatorv1.InstallModeSingletonHosted {
+		chartConfig.CreateNamespace = false
```
zhujian7, line 123: Should the noOperator be set to true in hosted mode?

pkg/bootstrap/render.go
```diff
@@ -398,29 +339,16 @@ func convertKubeConfigSecrets(ctx context.Context,
 			return nil, fmt.Errorf("kubeconfig key not found in secret %s", s.Name)
 		}
 
-		bootstrapKubeConfigSecrets = append(bootstrapKubeConfigSecrets, BootstrapKubeConfigSecret{
-			Name:       s.Name,
-			KubeConfig: base64.StdEncoding.EncodeToString(secret.Data["kubeconfig"]),
+		bootstrapKubeConfigSecrets = append(bootstrapKubeConfigSecrets, chart.BootStrapKubeConfig{
+			Name: s.Name,
+			//	KubeConfig: base64.StdEncoding.EncodeToString(secret.Data["kubeconfig"]),
```
zhujian7, line 344: delete? should this be encoded?

pkg/controller/hosted/controller.go
```diff
@@ -470,13 +473,23 @@ func createHostingManifestWork(managedClusterName string,
 		if err != nil {
 			panic(err)
 		}
+
+		unstructuredObj := &unstructured.Unstructured{}
+		if err := json.Unmarshal(jsonData, unstructuredObj); err != nil {
+			klog.Errorf("failed to unmarshal object %s: %v", string(jsonData), err)
+			continue
+		}
+		// the namespace has been created, should not be managed by manifestWork.
```
zhujian7, line 482: please also add some details why it should not be managed by manifestwork.
do we need to check the name of the namespace?

pkg/controller/hosted/controller.go
```diff
@@ -470,13 +473,23 @@ func createHostingManifestWork(managedClusterName string,
...
+		}
 		manifests = append(manifests, workv1.Manifest{
 			RawExtension: runtime.RawExtension{Raw: jsonData},
 		})
 	}
 
-	// For hosted mode, the klusterletManifestWork only contains a klusterlet CR
-	// and a bootstrap secret, delete it in foreground.
+	// For hosted mode, the klusterletManifestWork only contains a klusterlet CR, a bootstrap secret and an agent namespace,
+	// delete klusterlet and bootstrap in foreground, orphan the agent namespace.
```
zhujian7, line 492: It seems that the namespace is excluded above?

pkg/controller/importconfig/cluster_info.go
```diff
@@ -224,22 +224,18 @@ func buildImportSecret(ctx context.Context, clientHolder *helpers.ClientHolder,
 			WithManagedCluster(managedCluster).
 			WithKlusterletConfig(klusterletConfig).
 			WithPriorityClassName(priorityClassName)
-		yamlcontent, err = config.Generate(ctx, clientHolder)
+		yamlcontent, crdsV1YAML, err = config.Generate(ctx, clientHolder)
```
zhujian7, line 227: How many crds are left in the secret after the refactor? crd.yaml and crdv1.yaml? or only the crd.yaml?

pkg/bootstrap/render.go
haoqing0110: yes, we removed the limit for the hub components, which is a best practice for scaling. Is this line for klusterlet? it's fine to leave it for the spoke I think.

pkg/bootstrap/render.go
zhiweiyin318: CreateNamespace should be true. removed the comment.

pkg/bootstrap/render.go
zhiweiyin318: yes, this is for spoke cluster.

pkg/bootstrap/render.go
zhiweiyin318: yes,  noOperator is set  here https://github.com/stolostron/managedcluster-import-controller/pull/603/files#diff-34fd40905488f5ef71014d75f1d60e6a44679858fc341862946485ae03b11fddR183 there is a general function named installNoOperator  to set noOperator

pkg/bootstrap/render.go
zhiweiyin318: deleted. 
the data is not needed to encode here since the helm chart has encoded the data.

pkg/controller/hosted/controller.go
zhiweiyin318: added.
I don't think need to check the name of namespace here since this controller only handles the hosted mode.

pkg/controller/hosted/controller.go
zhiweiyin318: deleted the wrong comment.

pkg/controller/importconfig/cluster_info.go
zhiweiyin318: only crds.yaml now.  changed the name to crdsYAML.

xuezhaojun:
@zhiweiyin318 Is it possible we can also link the upstream PRs related to this story? In the Jira or in this PR's description.

zhiweiyin318:
@xuezhaojun  there is no related upstream PRs for this story. you can refer to the klusterlet helm chart and values https://github.com/open-cluster-management-io/ocm/tree/main/deploy/klusterlet/chart

PR #648, opened 2025-04-18
🌱 Remove dependency magnet

zhujian7:
The dependencymagnet is used to import the crdv1beta1, but now we do not need the crdv1beta1(removed by https://github.com/stolostron/managedcluster-import-controller/pull/603), and the upstream crdv1beta1 was also removed by https://github.com/open-cluster-management-io/api/pull/360, so this PR will delete the dependencymagnet


pkg/bootstrap/render.go
```diff
@@ -276,13 +276,16 @@ func (c *KlusterletManifestsConfig) Generate(ctx context.Context,
...
+			return nil, nil, fmt.Errorf("local secrets should be set")
+		}
 
 		c.chartConfig.Klusterlet.RegistrationConfiguration.FeatureGates = append(c.chartConfig.Klusterlet.RegistrationConfiguration.FeatureGates,
 			operatorv1.FeatureGate{
 				Feature: string(apifeature.MultipleHubs),
 				Mode:    operatorv1.FeatureGateModeTypeEnable,
 			})
-		c.chartConfig.Klusterlet.RegistrationConfiguration.BootstrapKubeConfigs = c.klusterletConfig.Spec.BootstrapKubeConfigs
+		c.chartConfig.Klusterlet.RegistrationConfiguration.BootstrapKubeConfigs = *c.klusterletConfig.Spec.BootstrapKubeConfigs.DeepCopy()
```
zhujian7, line 288: The LocalSecrets changed to be a pointer, now here we need to use DeepCopy, otherwise the ut will fail.

PR #649, opened 2025-04-21
remove the auto-import retry limit

elgnay:
The related issue: https://issues.redhat.com/browse/ACM-20121

### DIFF
--- a/pkg/controller/autoimport/autoimport_controller.go
+++ b/pkg/controller/autoimport/autoimport_controller.go
@@ -6,7 +6,6 @@ package autoimport
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"strings"
 
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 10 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -222,7 +222,7 @@ func TestReconcile(t *testing.T) {
 					},
 				},
 			},
-			expectedErr:             false,
+			expectedErr:             true,
```
zhujian7, line 225: I debugged this case locally, the output error is "kubeconfig is missing", should we consider to change the token in Line 220 to kubeconfig to make this case pass?

pkg/controller/autoimport/autoimport_controller.go
```diff
@@ -110,37 +109,6 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
...
-	totalRetry := 1
-	if len(autoImportSecret.Annotations[constants.AnnotationAutoImportCurrentRetry]) != 0 {
-		lastRetry, err = strconv.Atoi(autoImportSecret.Annotations[constants.AnnotationAutoImportCurrentRetry])
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-	}
-
-	if len(autoImportSecret.Data[constants.AutoImportRetryName]) != 0 {
-		totalRetry, err = strconv.Atoi(string(autoImportSecret.Data[constants.AutoImportRetryName]))
```
zhujian7, line 123: After the AutoImportRetryName is ignored, how to determinal if a cluster is imported failed from the web UI?

zhujian7:
@elgnay left 2 nit comments, and I did not see any configmap that you mentioned we need to determine whether it is a fresh install or an upgrade, that will be used in another PR?

pkg/controller/autoimport/autoimport_controller.go
elgnay: It can be determined by checking the `reason' of the condition `ManagedClusterImportSucceeded` is `ConditionReasonManagedClusterImportFailed` or not.

pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
elgnay: A positive case will be added in a separated PR.

elgnay:
> @elgnay left 2 nit comments, and I did not see any configmap that you mentioned we need to determine whether it is a fresh install or an upgrade, will that be used in another PR?

That's another PR.

PR #650, opened 2025-04-21
✨ support ClusterClaimConfiguration

zhujian7:
https://issues.redhat.com/browse/ACM-20141


pkg/bootstrap/render.go
```diff
@@ -46,6 +46,31 @@ var additionalClusterRoleFiles = []string{
...
+	"oauthredirecturis.openshift.io",
+	"openshiftversion-major",
+	"openshiftversion-major-minor",
+	"platform.open-cluster-management.io",
+	"product.open-cluster-management.io",
+	"schedulable.open-cluster-management.io",
+	"version.openshift.io",
+	"zero.hostedclustercount.hypershift.openshift.io",
+}
+
```
elgnay, line 73: It's unnecessary to list all of them. Just include the following suffixes:
- "hypershift.openshift.io"
- "open-cluster-management.io"
- "openshift.io"

zhujian7: There are 2 claims that I am not sure whether suffix is better:
```
	"exclude-from-backup.velero.io",
	"id.k8s.io",
```
or

```
	"velero.io",
	"k8s.io",
```

I don't know if customers need to use the `velero.io` and `k8s.io` as their claims' domain suffix.

PR #652, opened 2025-04-24
Support setting featuregates from klusterletConfig

qiujian16:


pkg/bootstrap/render.go
```diff
@@ -270,6 +271,19 @@ func (c *KlusterletManifestsConfig) Generate(ctx context.Context,
 		c.chartConfig.Images.ImageCredentials.DockerConfigJson = string(imagePullSecret.Data[corev1.DockerConfigJsonKey])
 	}
 
+	// feature gates
+	if c.klusterletConfig != nil {
+		for _, f := range c.klusterletConfig.Spec.FeatureGates {
+			if _, ok := apifeature.DefaultSpokeRegistrationFeatureGates[featuregate.Feature(f.Feature)]; ok {
+				c.chartConfig.Klusterlet.RegistrationConfiguration.FeatureGates = append(c.chartConfig.Klusterlet.RegistrationConfiguration.FeatureGates,
```
zhiweiyin318, line 278: multiHubs featureGate is appended below https://github.com/stolostron/managedcluster-import-controller/pull/652/files#diff-34fd40905488f5ef71014d75f1d60e6a44679858fc341862946485ae03b11fddL288-L292
there is an unusual case: if user disable multiHubs featureGate in klusterletConfig,  RegistrationConfiguration.FeatureGates  may have 2  multiHubs featureGates, one is true, another is false.

zhiweiyin318: Ignore the comment since the user should not set this incorrect configuration.

/lgtm

qiujian16: we can ignore the multiHubs featuregates here, but it is a bit weird. Or we do not add multiHubs in L302. WDYT @xuezhaojun

qiujian16: And updated the code with another option: check if multiplehubs feature gate is set already, only set if it is not.

PR #694, opened 2025-05-20
fix crd work is force deleted issue

zhiweiyin318:


pkg/controller/resourcecleanup/resourcecleanup_controller.go
```diff
@@ -216,6 +216,11 @@ func (r *ReconcileResourceCleanup) Cleanup(ctx context.Context, cluster *cluster
 	// only klusterletCRD manifestWork will be orphaned at the last, need to force delete.
 	klusterletCRDWorkName := fmt.Sprintf("%s-%s", cluster.Name, constants.KlusterletCRDsSuffix)
 	if len(works.Items) == 1 && works.Items[0].Name == klusterletCRDWorkName {
+		// the manifestWorks are deleted by registration controller.
+		// the agent may be orphaned if the CRD manifestWork is force deleted directly.
+		if works.Items[0].DeletionTimestamp.IsZero() {
+			return nil
```
elgnay, line 222: Should the managed cluster be re-queued after N seconds?

zhiweiyin318: will re-queued after 2 seconds 
https://github.com/stolostron/managedcluster-import-controller/blob/c1735bcb3ac6830de2b8d105cb2e74c56a8229d6/pkg/controller/resourcecleanup/resourcecleanup_controller.go#L90-L92

PR #696, opened 2025-05-23
support immediate-import annotation

elgnay:
The related issue: https://issues.redhat.com/browse/ACM-20816


pkg/helpers/helpers.go
```diff
@@ -327,6 +328,31 @@ func updateManagedClusterStatus(client client.Client, managedClusterName string,
...
+		return err
+	}
+
+	ImmediateImport, ok := managedCluster.Annotations[apiconstants.AnnotationImmediateImport]
+	if !ok || ImmediateImport == apiconstants.AnnotationValueImmediateImportCompleted {
+		return nil
+	}
+
+	managedCluster = managedCluster.DeepCopy()
+	managedCluster.Annotations[apiconstants.AnnotationImmediateImport] = apiconstants.AnnotationValueImmediateImportCompleted
```
qiujian16, line 344: does this mean that this annotation should not be set by an gitops tool.

elgnay: This annotation should not be set by a GitOps tool, as there's no valid reason for repeatedly triggering the import process of a managed cluster.

elgnay:
Force merge the PR and the fixing of the random e2e failure will be traced by a separated issue: https://issues.redhat.com/browse/ACM-21087

PR #700, opened 2025-05-28
Refactor ImportHelper: Remove applyResourcesFunc field and update related methods

xuezhaojun:
- Removed the applyResourcesFunc field from the ImportHelper struct.
- Updated the NewImportHelper function to reflect the removal of applyResourcesFunc.
- Renamed defaultApplyResourcesFunc to applyResourcesFunc and adjusted its usage in the Import method.

### DIFF
--- a/pkg/helpers/autoimport.go
+++ b/pkg/helpers/autoimport.go
@@ -87,10 +87,6 @@ func ContainInternalServerError(err error) bool {
 	return false
 }
 
-// ApplyResourcesFunc is a function to apply resources to the manged cluster to import it to the hub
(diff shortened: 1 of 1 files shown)


pkg/helpers/autoimport.go
```diff
@@ -102,12 +98,6 @@ type ImportHelper struct {
 	log            logr.Logger
 
 	generateClientHolderFunc GenerateClientHolderFunc
-	applyResourcesFunc       ApplyResourcesFunc
-}
-
-func (i *ImportHelper) WithApplyResourcesFunc(f ApplyResourcesFunc) *ImportHelper {
```
zhujian7, line 108: If I remember correctly, this was added for the unit tests, want to confirm if removing this will impact the tests?

PR #732, opened 2025-06-30
🌱 shorten the e2e consistently check time

zhujian7:


zhujian7:
did not change any critical thing, but the e2e still failed https://github.com/stolostron/managedcluster-import-controller/actions/runs/15968037829/job/45032655510?pr=732

PR #749, opened 2025-07-20
Reduce e2e test tmate timeout from 120 to 10 minutes

xuezhaojun:
This PR reduces the timeout for the tmate session in e2e tests from 120 minutes to 10 minutes.

The tmate session is used for debugging when e2e tests fail, and 10 minutes should be sufficient for most debugging scenarios while avoiding unnecessarily long-running workflows.

**Changes:**
- Updated `.github/workflows/e2e-tests.yml` to set `timeout-minutes: 10` instead of `timeout-minutes: 120`

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -41,4 +41,4 @@ jobs:
       - if: ${{ failure() }}
         name: Setup tmate session after Tests Failed
         uses: mxschmitt/action-tmate@v3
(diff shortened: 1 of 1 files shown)


xuezhaojun:
hi, @zhujian7 please take a look this PR, from my experience, if a dev wants to debug, he can extend the tmate timeout manually, but for normal case, we don't need a very long tmate timeout.

PR #750, opened 2025-07-22
add an annotation to prevent the cluster ns from being deleted

zhiweiyin318:
story: https://issues.redhat.com/browse/ACM-22336

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -19,6 +19,9 @@ const (
 	// keeping this secret after the cluster is imported successfully
 	AnnotationKeepingAutoImportSecret = "managedcluster-import-controller.open-cluster-management.io/keeping-auto-import-secret"
 
+	// AnnotationRemainNamespace is added to the ns by user to retain the namespace after the cluster is detached.
+	AnnotationRemainNamespace = "open-cluster-management.io/retain-namespace"
+
 	// LabelAutoImportRestore is the label key of auto import secret used for backup restore case
 	LabelAutoImportRestore = "cluster.open-cluster-management.io/restore-auto-import-secret"
 )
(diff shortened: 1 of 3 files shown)


pkg/controller/managedcluster/managedcluster_controller.go
```diff
@@ -93,18 +93,29 @@ func (r *ReconcileManagedCluster) Reconcile(ctx context.Context, request reconci
...
-
 	if !*modified {
 		return reconcile.Result{}, nil
 	}
 
+	if ns.Annotations == nil {
+		ns.Annotations = map[string]string{}
+	}
+	if _, ok := managedCluster.Annotations[constants.AnnotationRemainNamespace]; ok {
+		ns.Annotations[constants.AnnotationRemainNamespace] = ""
```
elgnay, line 114: Is there a specific reason why the annotation can be added to both the `ManagedCluster` and the cluster `Namespace`? In GitOps scenarios, if a user adds the annotation only to the Namespace and not the `ManagedCluster`, it could trigger an endless loop of annotation being added and removed on the Namespace.

zhiweiyin318: this feature requires the user to add the annotation to the cluster not the ns. 
import-controller will sync the annotation to the ns from the cluster because there is a case that the cluster is detached but there are other resources like clusterdeployment still in the ns, in this case need the annotation to determine if delete the ns after these resources are deleted.

elgnay: Why not just add the annotation on the cluster namespace instead of managedcluster?

zhiweiyin318: only add the annotation to the ns

PR #753, opened 2025-07-23
force delete crd work after deleting condition is true

zhiweiyin318:


pkg/controller/resourcecleanup/resourcecleanup_controller.go
```diff
@@ -218,9 +219,18 @@ func (r *ReconcileResourceCleanup) Cleanup(ctx context.Context, cluster *cluster
 	if len(works.Items) == 1 && works.Items[0].Name == klusterletCRDWorkName {
 		// the manifestWorks are deleted by registration controller.
 		// the agent may be orphaned if the CRD manifestWork is force deleted directly.
-		if works.Items[0].DeletionTimestamp.IsZero() {
+		crdWork := works.Items[0]
+		if crdWork.DeletionTimestamp.IsZero() {
+			return r.clientHolder.WorkClient.WorkV1().ManifestWorks(cluster.Name).
```
elgnay, line 224: I recall that the registration controller is responsible for deleting this ManifestWork. Why is the import-controller also deleting it?

zhiweiyin318: you are right. do not need delete again here. 
fixed.

pkg/controller/resourcecleanup/resourcecleanup_controller.go
```diff
@@ -218,9 +219,14 @@ func (r *ReconcileResourceCleanup) Cleanup(ctx context.Context, cluster *cluster
 	if len(works.Items) == 1 && works.Items[0].Name == klusterletCRDWorkName {
 		// the manifestWorks are deleted by registration controller.
 		// the agent may be orphaned if the CRD manifestWork is force deleted directly.
-		if works.Items[0].DeletionTimestamp.IsZero() {
+		// so the crd is deleting when the deleting condition of crd work is true
+		// and can force delete the crd work after the crd work is deleting.
+		crdWork := works.Items[0]
+		if crdWork.DeletionTimestamp.IsZero() ||
+			!meta.IsStatusConditionTrue(crdWork.Status.Conditions, workv1.WorkDeleting) {
```
elgnay, line 226: I believe `!meta.IsStatusConditionTrue(crdWork.Status.Conditions, workv1.WorkDeleting` is sufficient. What do you think?

PR #787, opened 2025-09-04
generate cluster import config secret

zhiweiyin318:
story: https://issues.redhat.com/browse/ACM-22961


qiujian16 (commented):
should we have a flag on importcontroller, or a setting in the related configmap? So we can disable adding the additional secret? Mainly on scalability concern.

PR #788, opened 2025-09-09
🌱 Remove duplicate params

zhujian7:


.tekton/managedcluster-import-controller-addon-mce-26-push.yaml
```diff
@@ -33,10 +33,6 @@ spec:
         - linux/arm64
         - linux/ppc64le
         - linux/s390x
-    - name: dockerfile
```
zhujian7, line 36: duplicated with line 26-29

PR #797, opened 2025-09-25
add grpc config into the bootstrap secret

zhiweiyin318:
story: https://issues.redhat.com/browse/ACM-23612


pkg/helpers/csr.go
```diff
@@ -9,6 +9,11 @@ import (
 
 var BootstrapSASuffix = "bootstrap-sa"
 
+const (
+	GRPCSAName      = "grpc-server-sa"
+	GRPCSANamespace = "open-cluster-management-hub"
```
skeeey, line 14: can we also use `hubNamespace` here to make this consistent

PR #1107, opened 2026-06-01
ACM-31624: re-import clusters after hub backup restore

rokej:
## Summary

- After hub backup/restore, Velero restores `ManagedCluster` status with `ManagedClusterImportSucceeded=True`, causing the autoimport controller to skip re-import under the default `ImportOnly` strategy
- Move the auto-import secret lookup and `restore-auto-import-secret` label check before the skip logic so backup-restore clusters are re-imported with the new hub URL
- Add unit test covering ImportOnly + already-imported + backup-restore secret scenario

Fixes [ACM-31624](https://redhat.atlassian.net/browse/ACM-31624)

## Test plan

- [x] `go test ./pkg/controller/autoimport/...`
- [ ] Restore hub with imported GKE/Azure clusters and verify they become Available

Made with [Cursor](https://cursor.com)

[ACM-31624]: https://redhat.atlassian.net/browse/ACM-31624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/controller/autoimport/autoimport_controller.go
+++ b/pkg/controller/autoimport/autoimport_controller.go
@@ -108,15 +108,6 @@ func (r *ReconcileAutoImport) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 	reqLogger.Info("Auto import strategy is fetched", "managedCluster", managedCluster.Name, "AutoImportStrategy", autoImportStrategy)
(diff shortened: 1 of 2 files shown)


rokej:
/hold

Pending Restore hub with imported GKE/Azure clusters and verify they become Available

rokej:
I cannot reproduce the problem with normal backup/restore flows. I am keeping the `/hold` label until the steps to reproduce the problem are known.

PR #1109, opened 2026-06-03
Fix DR restore race by marking ManifestWorks ReadOnly when disable-auto-import is set

tesshuflower:
## Summary

Fixes ACM-34619: Spoke cluster reconnects to backup hub instead of restore hub during DR restore (5-6% failure rate).

**Root cause:** Work-agent (on spoke) is still connected to the backup hub during DR restore. It periodically reconciles ManifestWorks every 4 minutes and overwrites the bootstrap secret that the restore hub just pushed to the spoke.  There is a manifestwork created by the managedcluster-import-controller that always exists that contains the bootsrap secret - so if this happens to be reconciled during the time window, this reset of bootstrap secret can happen (back to the original backup cluster).

**Solution:** When `disable-auto-import` annotation is set on ManagedCluster, mark all manifests in the klusterlet ManifestWorks as ReadOnly. This prevents work-agent from creating or updating resources during DR restore.

The manifestwork_controller in this repo works on 2 clusterlet manifestworks and these are the ones that will be affected by this change:

 1. `<cluster-name>-klusterlet-crds` - Contains the Klusterlet CRDs
  2. `<cluster-name>-klusterlet` - Contains the Klusterlet deployment and other resources (including the boostrap secret)

## Changes

1. **pkg/controller/manifestwork/manager.go**
   - Added watch trigger for disable-auto-import annotation changes
   - Reconciles ManifestWorks when annotation is added/removed

2. **pkg/controller/manifestwork/manifestwork_controller.go**
   - Added `buildManifestConfigs()` helper function
   - Generates ReadOnly ManifestConfigs when annotation is set
   - Applied to both klusterlet and klusterlet-crds ManifestWorks

3. **Unit tests** (manifestwork_controller_test.go)
   - Test: annotation set → ManifestWorks have ReadOnly configs
   - Test: no annotation → ManifestWorks have no ManifestConfigs

4. **E2E tests** (autoimport_test.go, e2e_suite_test.go)
   - assertManagedClusterManifestWorksReadOnly: verify ReadOnly configs exist
   - assertManagedClusterManifestWorksNotReadOnly: verify configs removed
   - Extended existing auto-import test to verify both states

## Test Plan

- [x] Unit tests pass
- [x] E2E tests compile
- [ ] E2E tests pass in CI
- [ ] Manual DR restore test with annotation set

## DR Restore Workflow

**Before restore (on backup hub):**
```bash
# Set annotation on the ManagedCluster resource on the BACKUP HUB
oc annotate managedcluster <cluster> import.open-cluster-management.io/disable-auto-import=""
```
This updates the ManifestWorks on the backup hub with ReadOnly configs. Work-agent (on spoke, connected to backup hub) receives the update via watch event.

**During restore:**
- Restore hub pushes bootstrap secret with restore hub URL to spoke
- Work-agent (still connected to backup hub) reconciles
- Work-agent sees ReadOnly configs from backup hub → does NOT overwrite bootstrap secret
- Bootstrap secret on spoke keeps restore hub URL ✅

**After spoke restarts:**
- Registration-agent uses bootstrap secret (has restore hub URL)
- Spoke connects to restore hub ✅
- Work-agent now connected to restore hub
- Backup hub decommissioned

## Related

- **Jira:** ACM-34619
- **Failure rate:** 5-6% (work-agent resync interval: 4 minutes ±50%)
- **Prevention:** ReadOnly strategy ensures work-agent only checks existence, cannot update

### DIFF
--- a/pkg/controller/manifestwork/manager.go
+++ b/pkg/controller/manifestwork/manager.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"strings"
 
+	apiconstants "github.com/stolostron/cluster-lifecycle-api/constants"
 	"github.com/stolostron/managedcluster-import-controller/pkg/constants"
 	"github.com/stolostron/managedcluster-import-controller/pkg/helpers"
 	"github.com/stolostron/managedcluster-import-controller/pkg/source"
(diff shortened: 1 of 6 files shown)


rokej:
/lgtm

If a user follows https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.16/html/business_continuity/business-cont-overview#prepare-new-hub to either stop or prepare the original backup hub cluster, this will work.

PR #1164, opened 2026-08-03
request to update ocm to 1.2 for mce 2.11

gparvin:
I received a request to update the ocm dependency to v1.2.1.  The update wanted go and linting to be updated as part of this.


rokej:
backplane-2.17 still has older `open-cluster-management.io/ocm v1.1.1-0.20260128054152-9d1a993e2c6f`. https://github.com/stolostron/managedcluster-import-controller/pull/1165 is the same PR for 2.17.

gparvin:
> Also, @gparvin Can you include a little more description about why this upgrade was needed?

This upgrade was requested by a managed service team as a result of one of their scans.  I did not see the scan details, just the request for the update which we tried to include you on.

